PR #7, opened 2020-08-22
Add bake_dictionary option

tmfink:
Originally, the dictionary at `src/resources/system.dic` needed to exist to compile the crate.
With this change, you can get the same behavior by enabling the `bake_dictionary` feature, but it is not enabled by default.
This makes compiles faster and will 

I documented this in the English README. The Japanese README still needs some translations--my 日本語 skills are too basic :sweat_smile:.

Other changes:
- Code formatting
- `Mode` enum: add documentation and `Copy` trait derive

Please let me know your thoughts--I'm open to suggestions.
Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 *.dic
+README*.html
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,11 +83,27 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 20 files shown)


mh-northlander (approved):
How to handle the dictionary is a big problem and we still need to consider, but I think bake-dictionary is a good option.

Documentation and tests seems no problem.

I understand that defining error enum is one of the error-handling best practices in rust.

Mode enum is small enough to be Copy.

PR #29, opened 2021-09-21
Fix: flush output for each lines

### LINKED ISSUE #22: Interactive usage does not work
Sudachi binary outputs nothing until you pass it end of file

mh-northlander:
fix #22


eiennohito:
LGTM
We should do it later probably only for Stdout, flushing buffer for real files can induce penalty by making more sysalls than needed.

PR #33, opened 2021-09-22
Remove lifetimes from Dictionary

eiennohito:
This is needed to use Pyo3.

Also, the new version of dictionary will use memory mapped files for loading dictionaries.


eiennohito:
What remains: replace old implementation with the new one (for the time being).
@mh-northlander  https://github.com/WorksApplications/sudachi.rs/pull/33/files#diff-a7009e0befab869ed21eca06cdf968ee88ae3bea5929f1f66d637d97e6652502R31-R47 Tokenizer now works without lifetimes (names are tentative).

eiennohito:
@mh-northlander Please squash commits into a single one when merging (or do only review, I will merge it afterwards itself).

mh-northlander (approved):

PR #39, opened 2021-09-27
Refactor Plugin Loading

### LINKED ISSUE #17: Move plugins to the main crate

### LINKED ISSUE #19: Make plugins to be built automatically without specifying --all
Can be done by specifying dependency as

plugin = { path = "path/to/plugin" }

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies

### LINKED ISSUE #20: Plugin loader: support built-in plugins
Use Java class names as identifiers of internal plugins

### LINKED ISSUE #21: Plugin-loader: Add variables
$bin - which resolves to the directory of the sudachi binary
$cfg - which resolves to the directory of the sudachi config (or config root?)

### LINKED ISSUE #24: Plugin loader: Refactor duplicate code

eiennohito:
Fixes #17 Fixes #19 Fixes #20 Fixes #21 Fixes #24

* Plugin loading logic is refactored, no repeated code now
* DSO plugins can now be resolved relative to executable with the syntax: `$exe/plugin` or relative to the configuration folder `$cfg/plugin`. 
* It is recommended not to specify platform dependent name parts for DSO plugins: `plugin` instead of `libplugin.so`. First name will correctly resolve on Windows/Linux/macOS.
* Built-in plugins are moved into the main crate, they can be specified in the configuration by their Java names (but that needs checking).

## Main design for the rework of plugin loader

We have core logic of plugin loading in a `PluginLoader` struct which is generic over implementations of `PluginCategory` trait.
`PluginCategory` captures plugin category specific logic: providing specific types, configuration extraction, and instantiation of bundled plugins.

PluginLoader is using fully qualified trait call syntax ([see previous paragraph before this link](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait)). It is required because type inference does not work if T is not present as one of arguments.


eiennohito:
Need to finish moving plugins into the main package to fix failing tests.

eiennohito:
This PR also should be squashed, I think.

eiennohito:
@mh-northlander ready for review & merging

mh-northlander (approved):

Issue #25, opened 2021-09-21
Python binding


mh-northlander:
Create [sudachipy](https://github.com/WorksApplications/SudachiPy) conpatible python binding

mh-northlander:
considering to use [PyO3](https://github.com/pyo3/pyo3), following hugging-face tokenizers

eiennohito:
What about Python API FFI like this? Main Idea - hide heavyweight API behind traits and use them via objects from Python side.

```rust
pub fn dictionary_from_config(cfg: Config) -> SudachiResult<Arc<dyn PyDictionary>> {
    todo!()
}

pub trait PyDictionary {
    fn tokenizer(&self) -> Box<dyn mut PyTokenize>;
}

pub trait PyTokenise {    
    fn tokenize(&mut self, input: &str, mode: Mode, debug: bool) -> SudachiResult<Vec<Morpheme>>;
}
```


Morpheme should be owning for Python, internal one would not work, binding should handle copying from Rust types to Python ones. Result also may be non-compatible.

mh-northlander:
PyO3 can only convert struct, so I'm considering to expose wrapper classes like following to hide rust API:
```rust
pub struct PyDictionary {
    inner: Arc<JapaneseDictionary>
}

impl PyDictionary {
    fn from_config(config: Config) -> Self {
        todo!();
    }

    fn tokenizer(&self) -> PyTokenizer{
        PyTokenizer { inner: self.inner }
    }
}

pub PyTokenizer {
    inner: StatelessTokenizer<Arc<JapaneseDictionary>>
}
```

eiennohito:
I'd like that we actually write down the list of requirements for the python binding.

PR #41, opened 2021-09-27
Python binding

mh-northlander:
See #25

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -1,0 +1,37 @@
+
+# sudachi.rs python
+
+This is the python binding of sudachi.rs.
+
+
+# Caution
+
+This project is under development and specifications may change drastically.
+
+
+# Setup
+
+1. Install python module `setuptools` and `setuptools-rust`.
+2. Run `python3 setup.py develop`.
+    - `develop` will create a debug build, while `install` will create a release build.
+3. Now you can import the module by `import sudachi.sudachi`.
+
+ref: [setuptools-rust](https://github.com/PyO3/setuptools-rust)
+
+
+# Example
+
+```python
+import sudachi.sudachi as ss
+
+dictionary = ss.Dictionary()
+tokenizer = dictionary.create()
+morphemes = tokenizer.tokenize("国会議事堂前駅")
+print(morphemes[0].surface())  # '国会議事堂前駅'
+print(morphemes[0].reading_form())  # 'コッカイギジドウマエエキ'
+print(morphemes[0].part_of_speech())  # ['名詞', '固有名詞', '一般', '*', '*', '*']
+
+tokenizer = dictionary.create(ss.SplitMode.A)
+morphemes = tokenizer.tokenize("国会議事堂前駅")
+print(list(map(str, morphemes)))  # ['国会', '議事', '堂', '前', '駅']
+```
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -4,16 +4,13 @@ pub mod dictionary;
 pub mod morpheme;
 pub mod tokenizer;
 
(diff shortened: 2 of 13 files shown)


python/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+    dictionary: Arc<JapaneseDictionary>,
+}
+
+#[pymethods]
+impl PyDictionary {
+    /// Creates a sudachi dictionary
+    #[new]
+    fn new(config_path: Option<PathBuf>, resource_dir: Option<PathBuf>) -> PyResult<Self> {
+        let config = Config::new(config_path, resource_dir, None)
+            .map_err(|e| PyException::new_err(format!("Error loding config: {}", e.to_string())))?;
```
eiennohito: typo: Error when loading config

sudachi/src/plugin/connect_cost.rs
```diff
@@ -23,7 +23,7 @@ use crate::dic::grammar::Grammar;
 use crate::prelude::*;
 
 /// Trait of plugin to edit connection cost in the grammar
-pub trait EditConnectionCostPlugin: Sync {
+pub trait EditConnectionCostPlugin: Sync + Send {
```
eiennohito: Can we merge adding `Send` to plugins before the full Python integration? I will need to propagate this change to plugin loader/holder rewrite as well.

eiennohito (commented):
Nice! 
We will deal with migration to StatefulAnalyzer later (when it will be implemented).

sudachi/src/plugin/connect_cost.rs
mh-northlander: Yes, I'll separate this part.

python/README.md
```diff
@@ -0,0 +1,37 @@
...
+    - `develop` will create a debug build, while `install` will create a release build.
+3. Now you can import the module by `import sudachi.sudachi`.
+
+ref: [setuptools-rust](https://github.com/PyO3/setuptools-rust)
+
+
+# Example
+
+```python
+import sudachi.sudachi as ss
```
eiennohito, line 25: double sudachi seems weird, but let's fix this later

python/src/tokenizer.rs
```diff
@@ -0,0 +1,131 @@
...
+            "B" | "b" => Ok(PySplitMode::B),
+            "C" | "c" => Ok(PySplitMode::C),
+            _ => Err("Mode must be one of \"A\", \"B\", or \"C\" (in lower or upper case)."),
+        }
+    }
+}
+
+#[pyclass(module = "sudachi.tokenizer", name = "Tokenizer")]
+pub struct PyTokenizer {
+    dictionary: Arc<JapaneseDictionary>,
```
eiennohito, line 84: does it need to actually hold the dictionary?

eiennohito (commented):
also please rebase on the develop branch

python/src/tokenizer.rs
mh-northlander: This is for `PyMorpheme`s, especially their POS.
Since we cannot clone Arc<dict> from `StatelessTokenizer`, we'll need to also keep dictionary.

Possible options are:
1. let `StatelessTokenizer.dict` pub
2. clone POS on `PyMorpheme` initialization

eiennohito: I see. Let's get rid of it then after #55, one of objectives of #55 is to reduce the number of references in Morphemes and make it possible to create lazy-loadable Morphemes on indexing.

eiennohito (approved):
LGTM, rebase & merge (this should be squashed)

PR #47, opened 2021-09-29
Change panics of tests to error checks

### LINKED ISSUE #45: Look into MeCab provider tests
Generally, we should not expect panics from test, we should check exact errors from them.

eiennohito:
fixes #45

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,6 +32,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.0.1"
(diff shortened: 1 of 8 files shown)


eiennohito:
@mh-northlander note that it is possible to use `&[u8]` as something that can be read from, extremely useful for tests
https://github.com/WorksApplications/sudachi.rs/pull/47/files#diff-24123d16ebca6e1760400eecba77fea338b0be2d55efd85f528003b44033e51eR66

mh-northlander (approved):

Issue #49, opened 2021-09-29
User-facing sentence splitter


eiennohito:
As per #28, we want to decouple analysis from sentence splitting.

This issue tracks creating sentence splitter as API.

PR #51, opened 2021-09-29
add prototype of sentence splitter public API

eiennohito:
also add split-only mode to the sudachi binary

NOT closing #49 because the API is probably not final and will need some rework after making tokenizer stateful


eiennohito:
@mh-northlander I think it's OK to merge it

mh-northlander (approved):

PR #57, opened 2021-10-01
Make CharacterCategory a bitset

### LINKED ISSUE #54: Make CharacterCategory a bitset
Part of #36 
Related to #53 

Rightnow it is a hashset of enums which is super inefficient compared to bitsets.

Making it a bitset will allow to make computational complexity of many algorithms to O(1) from O(n) as with hashmaps and remove allocations

eiennohito:
fixes #54

also cleaned up tests that were failing

### DIFF
--- a/sudachi/src/dic/category_type.rs
+++ b/sudachi/src/dic/category_type.rs
@@ -14,76 +14,154 @@
  * limitations under the License.
  */
 
-use std::collections::HashSet;
+use crate::error::SudachiError;
+use bitflags::bitflags;
 use std::str::FromStr;
 
-use crate::error::SudachiError;
+bitflags! {
+    /// A set of categories for a character
+    ///
+    /// Implemented as a bitset with fixed size
+    #[repr(transparent)]
+    pub struct CategoryType: u32 {
+        /** The fall back category. */
+        const DEFAULT = (1 << 0);
+        /** White spaces. */
+        const SPACE = (1 << 1);
+        /** CJKV ideographic characters. */
+        const KANJI = (1 << 2);
+        /** Symbols. */
+        const SYMBOL = (1 << 3);
+        /** Numerical characters. */
+        const NUMERIC = (1 << 4);
+        /** Latin alphabets. */
+        const ALPHA = (1 << 5);
+        /** Hiragana characters. */
+        const HIRAGANA = (1 << 6);
+        /** Katakana characters. */
+        const KATAKANA = (1 << 7);
+        /** Kanji numeric characters. */
+        const KANJINUMERIC = (1 << 8);
+        /** Greek alphabets. */
+        const GREEK = (1 << 9);
+        /** Cyrillic alphabets. */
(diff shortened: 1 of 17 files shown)


sudachi/src/dic/character_category.rs
```diff
@@ -46,25 +42,23 @@ pub enum Error {
...
+    categories: CategoryType,
 }
 
 /// CharacterCategory holds mapping from character to character category type
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub struct CharacterCategory {
-    ranges: Vec<Range>,
+    boundaries: Vec<u32>,
+    categories: Vec<CategoryType>,
```
mh-northlander, line 63: I feel it's better to have a comment about how boundaries and categories correspond to.
(we have `boundaries.len()+1` categories and `categories[i]` is for range `boundaries[i-1]..boundaries[i]`, right?)

sudachi/src/dic/category_type.rs
```diff
@@ -14,76 +14,154 @@
...
+use crate::error::SudachiError;
+use bitflags::bitflags;
 use std::str::FromStr;
 
-use crate::error::SudachiError;
+bitflags! {
+    /// A set of categories for a character
+    ///
+    /// Implemented as a bitset with fixed size
+    #[repr(transparent)]
```
mh-northlander, line 25: What is this attribute for? (I'm just curious.)

sudachi/src/input_text/utf8_input_text_builder.rs
```diff
@@ -152,13 +152,11 @@ impl Utf8InputTextBuilder<'_, '_> {
 
             // in rust, char corresponds to unicode scalar value
```
mh-northlander, line 153: I added this comment mentioning to java implementation.
Maybe it's ok to remove it.

mh-northlander (approved):

sudachi/src/input_text/utf8_input_text_builder.rs
eiennohito: OK

sudachi/src/dic/category_type.rs
eiennohito: From the binary point of view treat it as if it was a single u32, so this struct and related things actually do not exist in the binary at all, it's only single u32 which is usually passed via registers and does not even go to the stack most of the time.

eiennohito: In Rust this attribute can be assigned to any struct which has exactly one field.
https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent

sudachi/src/dic/character_category.rs
eiennohito: Certainly, will add.

sudachi/src/dic/category_type.rs
mh-northlander: I got it, thank you

Issue #55, opened 2021-09-30
AnalyzedSentence Design


eiennohito:
Prereq of #52 

We want a ponter-generic version of return container for the result of analysis.

Rust API wants references/mut references; Python binding wants Arc.

mh-northlander:
Considering to add `MorphemeList` following java/python:
```
struct MorphemeList<'a, T> {
    dictionary: T,
    input_text: &'a str,
    path: Vec<lattice::Node>
}
```

And let `Morpheme` have reference to `MorhemeList`:
```
struct Morpheme<'a, T> {
    list: &MorphemeList<'a, T>,
    index: usize,
    word_info: Option<WordInfo>, // for lazy load
}
```

`MorhemeList<'a, T>` will be returned from `tokenizer<T>.tokenize(input: &'a str)`.
We may change the struct for the path field later.

eiennohito:
In this formulation will it be usable in Pyo3? Even with wrappers.

mh-northlander:
For python side I considered to create a variant with no generic parameter:
```
struct PyMorphemeList {
    dictionary: Arc<JapaneseDictionary>,
    input_text: String,
    path: Vec<lattice::Node>
}

struct PyMorpheme {
    list: Arc<PyMorphemeList>,
    index: usize,
    word_info: Option<PyWordInfo>
}
```

eiennohito:
It would be nice if the Rust version of `MorphemeList` would be somehow reusable allocation-wise. At least in theory and not in the initial implementation.

For the Python that may be impossible (and Python FFI requires making copies for basically everything).

PR #65, opened 2021-10-05
Add morpheme_list

mh-northlander:
#55
Replace `Vec<Morpheme>` by `MorphemeList`.


mh-northlander:
Maybe it's better to let `Morpheme` have `Deref<trait morpheme-list>` instead of the reference to concrete `MorphemeList` type

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.path[index].end
+    }
+
+    fn get_surface(&self, index: usize) -> &String {
+        // returns substring of the original text which corresponds to the node at the given index
+        // we will need index mapping between original/modified input text
+        todo!();
+    }
+
+    fn get_word_info(&self, index: usize) -> WordInfo {
```
eiennohito: this should be `&WordInfo`

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    fn get_begin(&self, index: usize) -> usize {
+        self.path[index].begin
+    }
+
+    fn get_end(&self, index: usize) -> usize {
+        self.path[index].end
+    }
+
+    fn get_surface(&self, index: usize) -> &String {
```
eiennohito: this should be `&str`. Python wrapper should explicitly convert `&str` to PyString

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        self.index += 1;
+        Some(morpheme)
+    }
+}
+
+pub struct Morpheme<'a, T> {
+    list: &'a MorphemeList<'a, T>,
+    index: usize,
+    word_info: Option<crate::dic::lexicon::word_infos::WordInfo>,
```
eiennohito: this fully-qualified type should be in uses

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,213 @@
...
+{
+    pub fn begin(&self) -> usize {
+        self.list.get_begin(self.index)
+    }
+
+    pub fn end(&self) -> usize {
+        self.list.get_end(self.index)
+    }
+
+    pub fn surface(&self) -> &String {
```
eiennohito: `&str` to all strings

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,213 @@
...
+        &self.list.get_surface(self.index)
+    }
+
+    pub fn part_of_speech(&mut self) -> &Vec<String> {
```
eiennohito: `&Vec<T>` -> `&[T]` globally

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,213 @@
...
+        // returns substring of the original text which corresponds to the node at the given index
+        // we will need index mapping between original/modified input text
+        todo!();
+    }
+
+    fn get_word_info(&self, index: usize) -> WordInfo {
+        self.path[index].word_info.clone().unwrap()
+    }
+
+    fn split(&self, mode: Mode, index: usize, word_info: WordInfo) -> usize {
```
eiennohito: impossible to implement in the current contract

eiennohito (changes requested):

eiennohito:
Also, I didn't understand the comment about `Deref`.

mh-northlander:
> Also, I didn't understand the comment about `Deref`.

I'm wondering if we can reuse code in python bind part like:
```rust
// rust side
struct Morpheme<T> {
    list: T,
    index: usize,
}
impl <T> Morpheme where T: Deref, <T as Deref>::Target: MorphemeListTrait {
    /// 
}

// python side
impl MorphemeListTrait for PyMorphemeList { /// }

struct PyMorpheme {
    inner: Morpheme<Arc<PyMorphemeList>>
}
```

eiennohito:
I believe that Pyo3 does not allow to expose methods from trait implementations to Python, so that part can't go away.

Also, lazy loading WordInfo will require having mutable references to everything, which is not a good thing.

mh-northlander:
> I believe that Pyo3 does not allow to expose methods from trait implementations to Python, so that part can't go away.

Yes, anyway we need to write code to expose functions.
I leave this idea for later.

> Also, lazy loading WordInfo will require having mutable references to everything, which is not a good thing.

I faced problems around this and remove WordInfo field.

mh-northlander:
I changed the type of `MorphemeList.input_text` from `&str` to `String`.
I guess it's better to have `&str` from the view point of allocation, but failed to pass the reference to the original input text through analysis.

eiennohito:
It's OK to have the String there internally. In the memory reuse case we will swap it with the Utf8InputText buffer.

python/src/morpheme_list.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Returns the dictionary id which this word belongs
+    #[pyo3(text_signature = "($self)")]
+    fn dictionary_id(&self) -> i32 {
+        self.list.get_node(self.index).get_dictionary_id()
+    }
+
+    /// Returns the list of synonym group ids
+    #[pyo3(text_signature = "($self)")]
+    fn synonym_group_ids(&self) -> Vec<u32> {
```
eiennohito: &[u32]

sudachi/src/analysis/morpheme_list.rs
```diff
@@ -0,0 +1,275 @@
...
+        self.dict.grammar()
+    }
+}
+
+impl<T> MorphemeList<T>
+where
+    T: Deref + Clone,
+    <T as Deref>::Target: DictionaryAccess,
+{
+    pub fn split(&self, mode: Mode, index: usize) -> SudachiResult<MorphemeList<T>> {
```
eiennohito: As a general question, does it make sense to split only a single morpheme inside a path? (Except maybe tests)

Maybe the current Python API requires this, but we can provide better way to do things and deprecate bad ways later.
Usually we want to split the whole path at the same time, I believe.

sudachi/src/lib.rs
```diff
@@ -32,7 +32,7 @@ pub mod sentence_splitter;
 
 pub mod prelude {
     pub use crate::{
-        analysis::morpheme::Morpheme,
+        analysis::morpheme_list::MorphemeList,
```
eiennohito: I think there is no need to rename morpheme -> morpheme_list (because the morpheme is more domain-specific thing that the list of them).

sudachi/src/analysis/morpheme_list.rs
mh-northlander: I think it's possible that user want to change the split mode based on morpheme information (e.g. pos?).
I'll ask other people about this.

python/src/morpheme_list.rs
mh-northlander: pyo3 cannot automatically convert &[T] into a python list and we need to return owned vec.

ref: [pymethod constraint](https://pyo3.rs/latest/conversions/tables.html#:~:text=IntoPy%3CPyObject%3E%20is%20required%20for%20function%20return%20values.), [pyo3::IntoPy](https://pyo3.rs/main/doc/pyo3/conversion/trait.IntoPy.html#)

eiennohito: Then PyList directly at least (otherwise there are two conversions happening here).

sudachi/src/analysis/morpheme_list.rs
eiennohito: I think that that change is out of scope for this PR (because it is a change of API). Let's move it to another issue.

python/src/dictionary.rs
```diff
@@ -28,8 +28,9 @@ use crate::tokenizer::{PySplitMode, PyTokenizer};
 
 #[pyclass(module = "sudachi.dictionary", name = "Dictionary")]
 #[pyo3(text_signature = "(config_path, resource_dir)")]
+#[derive(Clone)]
 pub struct PyDictionary {
-    dictionary: Arc<JapaneseDictionary>,
+    pub dictionary: Arc<JapaneseDictionary>,
```
eiennohito: you probably meant pub(super)

eiennohito (approved):

PR #66, opened 2021-10-05
make trie lookup iterator-based

### LINKED ISSUE #63: Non-allocating Trie lookup

eiennohito:
fixes #63

### DIFF
--- a/sudachi/src/dic/lexicon_set.rs
+++ b/sudachi/src/dic/lexicon_set.rs
@@ -17,7 +17,7 @@
 use thiserror::Error;
 
 use crate::dic::lexicon::word_infos::WordInfo;
-use crate::dic::lexicon::Lexicon;
+use crate::dic::lexicon::{Lexicon, LexiconEntry, MAX_DICTIONARIES};
 use crate::prelude::*;
 
 /// Sudachi error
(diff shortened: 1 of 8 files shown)


eiennohito:
Gives ~5% global speedup

sudachi/src/dic/lexicon_set.rs
```diff
@@ -62,42 +59,38 @@ impl<'a> LexiconSet<'a> {
...
-                    .lookup(input, offset)?
-                    .iter()
-                    .map(|(wid, l)| self.build_dictword_id(did, *wid).map(|dwid| (dwid, *l)))
-                    .collect::<Result<Vec<_>, _>>()?,
-            );
-        }
-        vs.extend(self.lexicons[0].lookup(input, offset)?);
-
-        Ok(vs)
+    /// Searches dictionaries in the reverse order: user dictionaries first and then system dictionary
```
mh-northlander, line 83: Does the order affect the analysis result?

mh-northlander (approved):

sudachi/src/dic/lexicon_set.rs
eiennohito: It should not affect the analysis in any order (it is weight-based), except if two nodes have exactly same weights. I am not sure if user dictionary order and their effect on the analysis is specified somewhere.

PR #67, opened 2021-10-06
Make Utf8InputText stateful

### LINKED ISSUE #53: Stateful Utf8InputText
Part of #36 

Rework (and eventually replace) Utf8InputText/Utf8InputTextBuilder so they

1. Can be reusable between analyses
2. Should reuse allocated buffers for successive analyses
3. Should expose codepoint-based API for iterating modified string slices

### LINKED ISSUE #74: Enforce limit of 32k input bytes at once for tokenization
It will allow us to use u16 indices for words/codepoints internally.

Need to specially handle the (slightly unrealistic, but possible UB) case when normalization/rewriting extend it over 65k bytes.

eiennohito:
Fixes #53 Fixes #74 

It is now called InputBuffer

Its API is also made more rust-friendly.


eiennohito:
@mh-northlander OK, it's ready for pre-review.
I did not manage to replace current implementations yet. Also, OOV plugins are not tested (but integration tests pass).

sudachi/src/input_text/buffer/edit.rs
```diff
@@ -0,0 +1,372 @@
...
+pub fn resolve_edits(
+    source: &str,
+    source_mapping: &Vec<usize>,
+    target: &mut String,
+    target_mapping: &mut Vec<usize>,
+    edits: &mut Vec<ReplaceOp>,
+) -> usize {
+    let mut start: usize = 0;
+    let mut cur_len: isize = source.len() as isize;
+    for edit in edits.drain(..) {
```
mh-northlander, line 97: It seems that `edit` is assumed to be sorted. I want some comment about that (maybe it's better to add similar comment to the input_text plugin).

mh-northlander (approved):

Issue #52, opened 2021-09-29
SudachiPy compatibility


mh-northlander:
Part of #25.
Provide same interface to [SudachiPy](https://github.com/WorksApplications/SudachiPy) as much as possible.

Maybe after rust API become stable  (#28)

mh-northlander:
SudachiPy API to implement

- Dictionary
    - `__new__(config_path=None, resource_dir=None, dict_type=None)`
    - `create(self, mode=None) -> Tokenizer`
    - `close(self)`
- Tokenizer
    - `tokenize(self, text: str, mode=None, logger=None) -> MorphemeList`
- MorphemeList
    - behave as `List[Morpheme]`
        - `__getitem__`, `__len__`, `__iter__`
    - `__str__`
        - `"".join([morpheme.surface() for morpheme in self])`
    - `empty(cls) -> MorphemeList`
        - generates an empty MorphemeList
    - `get_internal_cost(self)`
        - returns the total cost of this path
    - `size(self)`
        - alias of `__len__`
- Morpheme
    - getter functions
        - `begin`, `end`
            - returns the index in the input text
        - `surface`, `part_of_speech`, `part_of_speech_id`
        - `dictionary_form`, `normalized_form`, `reading_form`
        - `is_oov`, `word_id`, `dictionary_id`, `synonym_group_ids`
    - `split(self, mode) -> MophemeList`
        - splits this morpheme based on mode
    - `get_word_info(self) -> WordInfo`
- WordInfo
    - getter
        - `surface`, `head_word_length`, `pos_id`
        - `normalized_form`, `dictionary_form_id`, `dictionary_form_id`, `reading_form`
        - `a_unit_split`, `b_unit_split`, `word_structure`, `synonym_group_ids`
    - `length(self)`
        - shorthand for head_word_length

eiennohito:
* Dictonary - OK
* Tokeniser - OK
* Morpheme / WordInfo - from the binding point of view they probably should be aliases if it is possible, maybe with lazy loading of dictionary information eventually
  * Otherwise there would be a lot of Arc internally in the morheme (or it is OK?)

mh-northlander:
Differences left:
`Dictionary.close()` is not implemented.
`Dictionary.__new__(-)` should take `dict_type` (#73)
`Tokenizer.tokenize(-)` takes `enable_debug` flag instead of `logger`.
`MorphemeList.empty(dict: Dictionary)` takes dictionary as args.
[...]

PR #77, opened 2021-10-08
Python compatibility

mh-northlander:
part of #52.

- Add `Dictionary.close()`
- Add `SplitMode` also under `Tokenizer`

### DIFF
--- a/python/src/tokenizer.rs
+++ b/python/src/tokenizer.rs
@@ -83,13 +83,17 @@ impl PyTokenizer {
     }
 
     /// Break text into morphemes
-    #[pyo3(text_signature = "($self, text, /, mode, enable_debug)")]
-    #[args(text, mode = "None", enable_debug = "None")]
+    ///
+    /// This ignores the logger provided
+    #[pyo3(text_signature = "($self, text, /, mode, logger, enable_debug)")]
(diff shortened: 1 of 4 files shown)


mh-northlander:
Should we let `Tokenizer.tokenize` to take `logger` (and do nothing with it)?

eiennohito (approved):

eiennohito:
Yes, accepting and ignoring logger parameter is OK.

python/src/tokenizer.rs
```diff
@@ -83,13 +83,17 @@ impl PyTokenizer {
...
+    #[pyo3(text_signature = "($self, text, /, mode, logger, enable_debug)")]
+    #[args(text, mode = "None", logger = "None", enable_debug = "None")]
+    #[allow(unused_variables)]
     fn tokenize(
         &self,
         text: &str,
         mode: Option<PySplitMode>,
-        enable_debug: Option<bool>, // want to take logger instead of debug flag
+        logger: Option<PyObject>,
+        enable_debug: Option<bool>,
```
eiennohito: I think that it is a bad idea to expose `enable_debug` here.

Current implementation of debug output is a placeholder, we should not expose it to Python.

eiennohito (approved):

Issue #56, opened 2021-09-30
Simplify import path for python binding


mh-northlander:
Part of #25.

With current python binding implementation we need `import sudachi.sudachi` to import module.
Add `__init__.py` to arrange it.

ref: Hugging Face Tokenizers seems have [a script](https://github.com/huggingface/tokenizers/blob/master/bindings/python/stub.py) to generate.

PR #78, opened 2021-10-08
Arrange python binding import path

mh-northlander:
#56.

We can import the module with `import sudachi`, instead of `import sudachi.sudachi`.

### DIFF
--- a/python/py_src/sudachi/__init__.py
+++ b/python/py_src/sudachi/__init__.py
@@ -1,0 +1,1 @@
+from .sudachi import *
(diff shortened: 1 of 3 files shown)


python/py_src/sudachi/__init__.py
```diff
@@ -0,0 +1 @@
+from .sudachi import *
```
eiennohito: It's better to enumerate everything explicitly here instead of * import

Explicit enumeration of API surface will be sort of documentation, or at least safeguards for us.

eiennohito (changes requested):

eiennohito (approved):

PR #80, opened 2021-10-11
Update readme: install path

mh-northlander:


eiennohito:
It's OK to commit readme changes directly, without going through MRs.

Issue #72, opened 2021-10-08
Add test for python binding


mh-northlander:
Part of #25.

eiennohito:
Somehow reusing test cases of SudachiPy would be the best solution, but we need only integration/API tests for Python bindings. And they should be written in Python.

PR #83, opened 2021-10-11
Add tests for python binding

mh-northlander:
#72.

tokenizer, dictionary and first 2 tests of word_info are copied from sudachipy.
SudachiPy passes additional tests.


eiennohito:
can you also add a bash script that will
1) create a new virtual environment (`python3 -m venv path`)
2) install python bindings there (in the development/editable mode)
3) run tests

eiennohito (approved):

PR #89, opened 2021-10-12
use lighter hasher for HashMaps

### LINKED ISSUE #88: Use lighter hasher for HashMaps
Rust's default hasher is too heavy for extremely small types, but as a good thing, we can completely customize it

eiennohito:
Use my hasher from Juman++, which was adopted from SeaHash, which was inspired by PCG pseudorandom generator

fixes #88

### DIFF
--- a/sudachi/src/dic/grammar.rs
+++ b/sudachi/src/dic/grammar.rs
@@ -19,7 +19,7 @@ use nom::{
     number::complete::{le_i16, le_u16},
 };
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


eiennohito:
Connection cost lookup and word param lookup will be replaced by better versions, but other uses of HashMap will remain.

eiennohito:
Found non-converted place, but it is fixed now
~33% speedup

mh-northlander (approved):

Issue #73, opened 2021-10-08
Read dictionary installed by `sudachidict_*`


mh-northlander:
Part of #25 and #52.

eiennohito:
Let's add label for python binding.

PR #93, opened 2021-10-13
Load sudachidict dictionary

mh-northlander:
#73

~WIP~: default dictionary path setting
SudachiPy: `dict_type` > `config.systemDict` > `sudachidict_core`.
sudachi.rs: arg > `config.systemDict` > baked dict.


mh-northlander:
Assume that "systemDict" key of default `sudachi.json` is an empty string as SudachiPy.
We will need to consider how to distribute resource files for python binding.

eiennohito:
Can we do path resolution completely in Python?
Then just pass the path to the dictionary to the Rust binding?
I am not sure all this logic should be written in Rust.

mh-northlander:
> Can we do path resolution completely in Python? Then just pass the path to the dictionary to the Rust binding? I am not sure all this logic should be written in Rust.

We can load python file using [`PyModule::from_code`](https://pyo3.rs/v0.14.3/python_from_rust.html#you-have-a-python-file-or-code-snippet-then-use-pymodulefrom_code).

In that case python side code will be like this:
https://github.com/WorksApplications/SudachiPy/blob/03d9ff8e7094859275984ee718e08c3fcefaeb6b/sudachipy/config.py#L42-L50

python/src/dictionary.rs
```diff
@@ -151,3 +153,21 @@ fn get_absolute_dict_path(py: Python, dict_type: DictionaryType) -> PyResult<Pat
 
     Ok(dict_path)
 }
+
+fn find_dict_path_py(py: Python, dict_type: &str) -> PyResult<PathBuf> {
+    let source_file = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
eiennohito: The path won't be like this in the installed module (after doing `pip install`).

I think it is OK to do this resolution inside `__init__.py`, or even put the resolution function there (prefixed with underscore so it will be sort of private for the module).

python/py_src/sudachi/dictionary_path.py
```diff
@@ -0,0 +1,21 @@
...
+    return str(dic_path.absolute())
+
+
+def find_dict_path(dict_type='core'):
+    is_installed = find_spec('sudachidict_{}'.format(dict_type))
+    if is_installed:
+        return get_absolute_dict_path(dict_type)
+    else:
+        raise ModuleNotFoundError(
+            'Package `sudachidict_{}` dose not exist. '
```
eiennohito: typo: does not exist

python/src/dictionary.rs
mh-northlander: like this?
```rust
let module = PyModule::import(py, "sudachi.dictionary_path")?;
```
In this case I think it's better to separate file, not to load other functions.

eiennohito: What do you mean by load?
Resolving the path of the pip-installed dictionary is a core functionality of the Python binding, as long as it is shown as "do not call" for users, e.g. having `_revolve_dic_internal`, it should not be a problem. Python does not have the concept of visibility.

eiennohito: `PyModule.import` should work though

mh-northlander: I meant the loading cost of `PyModule::import`, which will loads entire module.
If we put functions in `__init__.py`, maybe `PyModule::from_code(py, "from sudachi import function")`, is better.

eiennohito: Creating a new dictionary is one off operation, it is OK for it to be more or less heavy.

eiennohito (approved):

PR #94, opened 2021-10-13
Split Node into smaller things

### LINKED ISSUE #50: Stateful Tokeniser
Part of #36 

It should reuse allocations between analyses.
Parts:
* #81
* #71

### LINKED ISSUE #71: Decrease size of Node
Nodes are copied around a lot and Rust does not have placement semantic yet.
The easiest way to improve the analysis speed is to decrease the size of the Node structure.

It should be made **at most 32 bytes** (currently 264 bytes).
It means that most of the fields must go away and WordInfo must be moved out of the structure.

### LINKED ISSUE #81: Make Lattice stateful
It should be able to reuse allocations between analyses

eiennohito:
Fixes #71, Fixes #81, Fixes #50

Split Node into:

* Node - produced by dictionary lookup (fast) and OOV
* V(iterbi)Node - used in analysis
* ResultNode - combination of first two plus WordInfo

Node always uses modified/rewritten indices
ResultNode uses modified indices during the path rewrite and original indices afterwards. It also holds both codepoint and byte indices to the data

WordId output changed to (dic_id, word_id).
dic_id is -1 for OOV only in output, internal value is 15=0xf.
This must be ported to Java.


eiennohito:
@mh-northlander I will add more documentation comments and check for unused InputBuffer methods, but this is almost ready.

python/src/morpheme.rs
```diff
@@ -14,40 +14,24 @@
...
-    type Target = MorphemeList<Arc<JapaneseDictionary>>;
-    fn deref(&self) -> &Self::Target {
-        &self.list
-    }
-}
+type PyMorphemeList = MorphemeList<Arc<JapaneseDictionary>>;
 
 /// A list of morphemes
-#[pyclass(module = "sudachi.morphemelist", name = "MorphemeList")]
+#[pyclass(module = "sudachi.morpheme", name = "MorphemeList")]
```
mh-northlander: this should be `sudachi.morphemelist` (for SudachiPy compatibility).

sudachi/src/input_text/buffer/mod.rs
```diff
@@ -62,7 +62,8 @@ pub struct InputBuffer {
     /// Byte mapping from normalized data to originals.
     /// Only values lying on codepoint boundaries are correct. Byte-based indexing.
     m2o: Vec<usize>,
-    /// Buffer for normalization
+    /// Buffer for normalization.
+    /// After building it is used as c2b for original data.
```
mh-northlander: maybe this is `b2c`, based on l.200 `fn fill_orig_b2c`?

mh-northlander (changes requested):

python/src/morpheme.rs
eiennohito: Rebase fail.
But Tests didn't caught that.

sudachi/src/input_text/buffer/mod.rs
eiennohito: Thanks for the catch

python/src/morpheme.rs
eiennohito: Also, it worked as is in test.
May be related to https://pyo3.rs/v0.14.5/module.html#python-submodules

eiennohito:
Ok, done, also added an import test for Python.

mh-northlander (approved):

python/src/morpheme.rs
mh-northlander: This is actually kind of comment.
https://pyo3.rs/v0.14.5/class.html#customizing-the-class

Issue #84, opened 2021-10-11
Fix import path for python binding


mh-northlander:
part of #52.

Currently all components locate at the root module. Change import path same as SudachiPy.
i.e. `from sudachi import Dictionary` -> `from sudachipy.dictionary import Dictionary`

PR #96, opened 2021-10-14
Add sudachipy like import path

mh-northlander:
#84.

We can still use direct import path like `from sudachi import Dictionary`

### DIFF
--- a/python/src/lib.rs
+++ b/python/src/lib.rs
@@ -15,46 +15,15 @@
  */
 
 use pyo3::prelude::*;
-use pyo3::wrap_pymodule;
 
 mod dictionary;
 mod morpheme;
 mod tokenizer;
 mod word_info;
 
-#[pymodule]
-fn dictionary(_py: Python, m: &PyModule) -> PyResult<()> {
-    m.add_class::<dictionary::PyDictionary>()?;
-    Ok(())
-}
-
-#[pymodule]
-fn tokenizer(_py: Python, m: &PyModule) -> PyResult<()> {
-    m.add_class::<tokenizer::PySplitMode>()?;
-    m.add_class::<tokenizer::PyTokenizer>()?;
-    Ok(())
-}
-
-#[pymodule]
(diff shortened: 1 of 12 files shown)


python/src/lib.rs
```diff
@@ -15,15 +15,46 @@
...
-pub mod morpheme;
-pub mod tokenizer;
-pub mod word_info;
+mod dictionary;
+mod morpheme;
+mod tokenizer;
+mod word_info;
+
+#[pymodule]
+fn dictionary(_py: Python, m: &PyModule) -> PyResult<()> {
```
eiennohito: Do we really need them exported as modules into Python via FFI?
Faking that purely in python will remove FFI entities which is a good thing.

mh-northlander: Right, I'll change it

python/tests/test_dictionary.py
```diff
@@ -15,7 +15,8 @@
 import os
 import unittest
 
-from sudachi import Dictionary, Tokenizer
```
eiennohito: I think that original imports were better, we are promoting new ways of import as the correct usage, am I correct?

But we still need to have a test that old ways of import do not break (a new one, which only does imports inside test methods).

eiennohito (changes requested):

python/tests/test_dictionary.py
mh-northlander: > I think that original imports were better, we are promoting new ways of import as the correct usage, am I correct?

Currently we don't have concrete plan about this.
I personally think we can put everything in the root module since the amount is not so large.

In the future we will change the API along with the rust API and may want a different module structure.
For that time, I guess we can keep current (old) path as default.

eiennohito: > I personally think we can put everything in the root module since the amount is not so large.

Completely agree with that, and it is also easier to use it that way.

I would think that we should move to having the public API under the main package (`sudachi`) by default and keep the old modules only for compatibility (without deprecating them though).

eiennohito (approved):

Issue #68, opened 2021-10-07
Change Python API


mh-northlander:
relates to #52.

Some of the current python APIs seems not necessary (e.g. `MorphemeList.split`).
We may provide better way to do.

mh-northlander:
API to remove:

- `MorphemeList.empty`
  - `MorphemeList` (and `Morpheme`) should be provided via `Tokenizer` and users should not generate them by themselves.
- `Morpheme.get_word_info`
  - `WordInfo` is too raw to expose. We provide necessary data via `Morpheme`.
- `Morpheme.split()`
  - API around this will be change (#92)

add deprecated attribute

mh-northlander:
Close this for now, since we don't have a plan for additional changes in the first release.

PR #97, opened 2021-10-14
Add deprecation warning

mh-northlander:
#68.

### DIFF
--- a/python/src/morpheme.rs
+++ b/python/src/morpheme.rs
@@ -58,13 +58,16 @@ impl PyMorphemeListWrapper {
     /// Returns an empty morpheme list with dictionary
     #[classmethod]
     #[pyo3(text_signature = "(dict)")]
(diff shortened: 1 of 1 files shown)


eiennohito:
Does this gets printed only once or on every usage?

mh-northlander:
This uses python's warning system.
By default, only the first occurrence will be printed if it is used in `__main__`.

ref: https://docs.python.org/3/library/warnings.html#:~:text=category%20for%20warn().-,DeprecationWarning,-Base%20category%20for

eiennohito (approved):

eiennohito:
For reference: [PyErr::warn](https://pyo3.rs/main/doc/pyo3/struct.PyErr.html#method.warn) calls [PyErr_WarnEx](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WarnEx).

PR #103, opened 2021-10-18
Build sdist

### LINKED ISSUE #100: Default resource and config files for python binding
Distribute resource files (except dictionaries) and the config file.

We also need to modify code to load them.

mh-northlander:
resolve #100, and update files to build sdist bundle

Use `build-sdist.sh` instead of running `python setup.py sdist`.


eiennohito:
Is it possible to somehow reduce resource duplication?

mh-northlander:
> Is it possible to somehow reduce resource duplication?

`python/py_src/sudachi/resource` is now used as the default config/resource and necessary to make `python setup.py develop` work.

For the sdist, we can remove them by adding data copy to `build-sdist.sh`.
Also note that `systemDict`'s value of `sudachi.json` is empty to mimic SudachiPy's config behavior.

eiennohito:
Noted on differences, please document it somewhere.

eiennohito (approved):

PR #107, opened 2021-10-19
Add docs-deploy workflow

### LINKED ISSUE #106: Add CI to deploy python doc to github pages
for now this only targets to deploy `python/docs`.

mh-northlander:
resolves #106

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -1,0 +1,33 @@
+name: Deploy Documentation
+# deploy python doc only for now
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


eiennohito:
It's better to do this together with generating documentation for Rust (just because we can) or at least make sure that URLs will NOT break after adding the Rust documentation later.

eiennohito:
e.g. publish Python docs to `/python` and Rust docs to `/rust` and have a html file at root that will redirect to both locations.

eiennohito:
Building rust docs: `cargo doc --no-deps`

Issue #104, opened 2021-10-18
Create Binary Distributions


mh-northlander:

eiennohito:
1. Binary build should use PGO: https://doc.rust-lang.org/rustc/profile-guided-optimization.html
2. What Python versions we are going to cover? 3.6+? https://endoflife.date/python Should we consult with GinZa for this? @hiroshi-matsuda-rit
3. What platforms should we build binaries? Building for Linux is easy-ish (https://github.com/pypa/manylinux). See also https://github.com/marketplace/actions/python-wheels-manylinux-build (or other Google results).  For Windows/maxOS I am not sure.

eiennohito:
For PGO we can't use available actions because we have to invoke `python setup.py` in a special manner (with RUSTFLAGS environment variable)

eiennohito:
Current plan is to build binaries for 
Python versions: 3.6, 3.7, 3.8. 3.9 and 3.10 
OS: Linux (with manylinux containers), Windows, macOS
Linux - do only x86_64 builds with PGO
Windows - do only x86_64 builds without PGO (or maybe with if it will be easy)
macOS - do universal x86_64/Aarch64 builds (!) without PGO (or similarily to Windows)

PR #111, opened 2021-10-20
Add workflow to release python

### LINKED ISSUE #105: Add CI to deploy python binding

mh-northlander:
Closes #105.

`bdist` part is not yet included (#104).

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -1,0 +1,49 @@
+name: Python Release
+
+on:
+  push:
+    tags:
+      - python-v*
+
+jobs:
+  build-sdist:
+    name: Build source distribution
+    runs-on: Ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Install dependencies
+        run: python3 -m pip install -U setuptools setuptools-rust
+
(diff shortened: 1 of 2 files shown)


.github/workflows/python-release.yml
```diff
@@ -0,0 +1,49 @@
+name: Python Release
+
+on:
+  push:
+    tags:
+      - python-v*
```
eiennohito: Let's use `v*` pattern here

.github/workflows/python-release.yml
```diff
@@ -0,0 +1,49 @@
+name: Python Release
```
eiennohito: Name -> Python Source Package (Common)

eiennohito (changes requested):
Also, file should be `release-python-source.yml`

.github/workflows/python-release.yml
mh-northlander: You mean we use a suffix to distinguish rust/python like `v0.1.0-python`?

eiennohito: I don't think that we should distinguish them (Rust vs Python) at all. They should be completely in-sync to reduce mental load of version comparison.

eiennohito: When there will be the need for distinguishing, I hope the SudachiPy repo will already be free for use for the Python bindings.

eiennohito (approved):

PR #114, opened 2021-10-20
Add cli tool for python

### LINKED ISSUE #101: Add cli binary to python binding
Distribute a CLI tool with python binding.

mh-northlander:
Closes #101 ~~and #115~~.

Base code is copied from SudachiPy.


.github/workflows/python.yml
```diff
@@ -0,0 +1,35 @@
...
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macOS-latest ]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download dictionary
+        run: bash fetch_dictionary.sh
+
```
eiennohito: please add caching here from Rust build

eiennohito (commented):
Doing output via logger is super strange, I don't like it.

Logging (diagnostic output by definition) should be different from the main output (main purpose of this program) and should use different facilities.

Please divide this PR into CI integration (which is good) and executable only (which is not good and should be cleaned-up).

eiennohito (approved):

PR #119, opened 2021-10-21
fix NonBreakChecker

### LINKED ISSUE #79: Fix NonBreakChecker
Sentence splitter originally checked if a morpheme could pass the possible sentence boundary and forbid split on boundaries that could span a morpheme. This check is now not functional

eiennohito:
fixes #79
It does not check can_bow (that check was probably redundant)
and has decreased detection range (~10 characters) from 64 bytes

### DIFF
--- a/sudachi/src/sentence_detector.rs
+++ b/sudachi/src/sentence_detector.rs
@@ -32,22 +32,17 @@ impl<'a> NonBreakChecker<'a> {
 }
 impl NonBreakChecker<'_> {
     /// Returns whether there is a word that crosses the boundary
-    /// TODO: this implementation is broken now
+
     fn has_non_break_word(&self, input: &str, length: usize) -> SudachiResult<bool> {
         // assume that SentenceDetector::get_eos called with self.input[self.bos..]
         let eos_byte = self.bos + length;
         let input_bytes = input.as_bytes();
-        const LOOKUP_BYTE_LENGTH: usize = 64;
+        const LOOKUP_BYTE_LENGTH: usize = 10 * 3; // 10 Japanese characters in UTF-8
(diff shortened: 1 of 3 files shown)


sudachi/src/sentence_detector.rs
```diff
@@ -32,22 +32,17 @@ impl<'a> NonBreakChecker<'a> {
...
         for i in lookup_start..eos_byte {
-            // TODO: fixme
-            // if !input.can_bow(i) {
-            //     continue;
-            // }
             for entry in self.lexicon.lookup(input_bytes, i) {
                 let end_byte = entry.end;
-                let char_count = input[i..end_byte].chars().count();
-                if end_byte > eos_byte || (end_byte == eos_byte && char_count > 1) {
+                if end_byte >= eos_byte {
```
mh-northlander: This needs to be `>`, since words that end at `eos_byte` are not "non-break-word".

I guess the another condition `end_byte == eos_byte && char_count > 1` ([java impl](https://github.com/WorksApplications/Sudachi/blob/9a02c6bbae6338274198deb59d9bdddf0eae328f/src/main/java/com/worksap/nlp/sudachi/JapaneseTokenizer.java#L293)) is to prevent splitting just after words like "モーニング娘。" which end with eos-like char.
It would be better to keep this part too.

mh-northlander (changes requested):

sudachi/src/sentence_detector.rs
eiennohito: Iterating over all characters is also not the best thing to do here. 

I've added the iteration limit and added some comments for the future

mh-northlander (approved):

PR #132, opened 2021-10-22
Merge all build scripts into a single uberscript

eiennohito:
It builds both source and binary packages and only then uploads everything to PyPI only if all builds succeed.

All builds are also accessible as build artifacts.


eiennohito:
squash merge

mh-northlander (approved):

PR #134, opened 2021-10-22
Add script to update version

### LINKED ISSUE #128: Automate version management
Updating the version of all of rust/python/etc. is too hard for human.

mh-northlander:
Closes #128.

Readme is out of scope.

### DIFF
--- a/update_version.sh
+++ b/update_version.sh
@@ -40,3 +40,16 @@ sed -i -r "1,/^ *version=/ s/^ *version=\"${VERSION_FROM}\",$/    version=\"${VE
 PYDOC_CONF_FILE="./python/docs/source/conf.py"
 echo $PYDOC_CONF_FILE
 sed -i -r "1,/^release = '/ s/^release = '${VERSION_FROM}'$/release = '${VERSION_TO}'/" $PYDOC_CONF_FILE
+
+
+# check
+echo ""
+echo "grep with previous version number:"
+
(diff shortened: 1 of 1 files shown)


eiennohito (commented):
Cargo.toml updater possibly can update dependency versions.

I think that it's better to be super foolproof here and specialize updaters for each file type.
For example, in `Cargo.toml` change exactly `^version = \"cur\"` (for dependencies it is super rare for version to be at the beginning of the line) to the next supplied version (not with regex captures, but hardcoding both versions in the sed expression). Also it is possible to check if sed actually does the replacement: https://stackoverflow.com/a/66365284/976247

Finally, it would be good to grep for the old version in the end of the script to make sure that nothing was left out.

update_version.sh
```diff
@@ -17,22 +17,39 @@ fi
...
+sed -i -r "1,/^release = '/ s/^release = '${VERSION_FROM}'$/release = '${VERSION_TO}'/" $PYDOC_CONF_FILE
+
+
+# check
+echo ""
+echo "grep with previous version number:"
+
+set +e # allow grep to exit with 1 (no line match)
+
+FILES="$CARGO_FILES $PY_SETUP_FILE $PYDOC_CONF_FILE"
```
eiennohito: I think that it is OK to grep everything, it is for human to decide whether something was skipped or not.

eiennohito (approved):
Otherwise, LGTM

PR #143, opened 2021-10-25
Build User and System Dictionary

### LINKED ISSUE #13: Add build-dictionary mode
Current sudachi.rs cannot build both system and user dictionary.
We want to a cli tool for that feature as java/python

We can still use dictionaries build by java/python version

eiennohito:
Fixes #13 

Add build dictionary functionality to Sudachi.rs.

Implementation is done in library fashion to allow easy inclusion in the Python bindings.
Python bindings for dictionary building themselves are out of scope of this PR and will be done afterwards.

The feature should be relatively well-tested.

I also added the functionality to dump the dictionary contents and compared the dumps with the dictionaries from the Java version. Dumps are identical, but the dictionaries themselves differ in size because of different implementation of trie creation.
This PR uses https://github.com/takuyaa/yada for trie creation and tries themselves differ from the Java version, but are compatible.


eiennohito:
@mh-northlander cli is not tested, everything else should be ready for review

sudachi-cli/src/build.rs
```diff
@@ -0,0 +1,266 @@
...
+            part.size(),
+            unit,
+            part.time().as_secs_f32()
+        )
+    }
+}
+
+fn dump_part(dict: PathBuf, part: String, output: PathBuf) {
+    let cfg = Config::new(None, None, Some(dict)).unwrap();
+    let dict = JapaneseDictionary::from_cfg(&cfg).unwrap();
```
mh-northlander: I guess this fails to load an user dictionary due to the header version check

mh-northlander (commented):
lib part seems no problem for me.

sudachi-cli/src/build.rs
eiennohito: User dictionaries should be built for the system dictionary, or am I understanding something incorrectly?

eiennohito: Nevermind the previous comment, this is in the dump logic which should be able to handle both types

PR #149, opened 2021-10-26
Add sudachipy.pyi

### LINKED ISSUE #147: Provide type hints for Python API

mh-northlander:
Fixes #147 

Add pyi file for type hints. `sudachipy.pyi` is enough to cover all classes.
I used `mypy`'s `stubgen` cli tool to generate base of pyi file.

Also fix type signatures. The annotation for positional-only parameters was wrong.
ref: [PEP570](https://www.python.org/dev/peps/pep-0570)


eiennohito:
Some questions: 
1. How difficult is to sync .pyi with the Rust code? E.g. when adding new functions.
2. IDEs usually show documentations from Python docstrings, how .pyi interact with Rust comments? Are they shown in IDEs? (Checking both PyCharm and VS Code is sufficient).

mh-northlander:
> 1. How difficult is to sync .pyi with the Rust code? E.g. when adding new functions.

Currently it is handmade. We need to update pyi manually.

> 2. IDEs usually show documentations from Python docstrings, how .pyi interact with Rust comments? Are they shown in IDEs? (Checking both PyCharm and VS Code is sufficient).

a. PyCharm
  Editor shows rust comments (I guess it shows `__doc__`). If we add comments to pyi, it overwrites rust comments.

b. VSCode
  Rust comments are not visible from the editor. If we add comments to pyi, they're shown.

eiennohito:
OK, let's go the manual path until the API surface becomes too large to support.

So, Rust comments on the API should be reflected in .pyi file, probably copying them verbatim will do.

eiennohito (approved):

PR #153, opened 2021-11-08
add Python binding for building user/system dictionaries

### LINKED ISSUE #152: Expose dictionary build from Python bindings
After #143 is merged

eiennohito:
Fixes #152


python/py_src/sudachipy/command_line.py
```diff
@@ -17,9 +17,11 @@
 import logging
 import os
 import sys
+from pathlib import Path
 
 from . import __version__
 from . import Dictionary, SplitMode
+from . import sudachipy
```
mh-northlander, line 24: How about to re-import `build_system_dic` and `build_user_dic` in `sudachipy/__init__.py` so that we can import them by `sudachipy.build_~` (instead of `sudachipy.sudachipy.build_~`)?

mh-northlander (approved):
we also need to update documents and pyi file

python/py_src/sudachipy/command_line.py
eiennohito: I assumed that we do not expose them (maybe yet) as public API, so didn't do that. Should we expose them then?

mh-northlander: I guess it's better to provide API at some point, for the use case like preprocess data and build dictionary in a single script.
But if you have some plan to change API around dictionary, it is ok to postpone this.

eiennohito: It is sure useful to provide ways of building dictionaries as API, but I am not so sure about the current form (which is certainly usable though).

I will make this an issue for the future work.

PR #167, opened 2021-11-15
Add HuggingFace-compatible PreTokenizer

### LINKED ISSUE #38: Huggingface Tokenisers Integration
Probably, the easiest way to the integration would be to use CustomPretokeniser hook in the python bindings.
https://github.com/huggingface/tokenizers/blob/master/bindings/python/src/pre_tokenizers.rs#L547

Good things:
* Python will handle dynamic library loading for us

Bad things:
* Some performance overhead on python method calls
* Some performance overhead on string copying

Anything else would require significant additions into the Huggingface Tokeniser framework.
Let's prototype this integration and maybe switch to another one if the overhead of this one will be too high.

Plan:

Add `create_pretokenizer()` method of Dictionary class (Python Binding) which will return an object which is compatible with HuggingFace.

Depends on #166

### LINKED ISSUE #166: Read subset of WordInfo fields
Because it is not needed.

Plugins are modifying this information, so let's try to explore how to implement both reading and not reading dictionary information at the same time

### LINKED ISSUE #168: Use tokenization-only mode in PreTokenizer
Depends on #166

eiennohito:
Depends on PR #176 (includes its comments)

Fixes #38 Fixes #166 #Fixes #168

Created as `Dictionary.pre_tokenizer([mode])`, where the default mode is C.

Implementation can be used from multiple threads, with a single PreTokenizer instance shared between all threads. It will create thread local Sudachi tokenizer and use it to perform the actual tokenization. Also, implementation releases GIL while doing the analysis, so it should achieve some speedup when used multithreaded.


mh-northlander (approved):

eiennohito:
Added custom handler

eiennohito:
rebased on develop branch

PR #174, opened 2021-11-22
precompute POS data for Python

eiennohito:
as a side effect, POS are now Tuples instead of Lists


mh-northlander (approved):
Are there any limitation or performance issue for using PyList?
(Anyway I think tuple is a better container for POS)

eiennohito:
When considering shared instances it's not performance, but correctness issue.
Because lists are mutable, it is possible to add and remove elements from them. 
It's better to be on the safe side here.

PR #180, opened 2021-11-24
Add POS matcher

### LINKED ISSUE #177: Python API: POS matcher
Implement a way to match a morpheme to a set of POS tags.

Need to decide an interface.

Callback interface which uses a callable to check eligible POS during creation.

```python
matcher = dic.pos_matcher(lambda p: p[0] == "名詞")
```

Declarative interface which expects user to pass a list of POS tuples.

```python
matcher = dic.pos_matcher([('名詞', '一般', '*', '*', '*', '*')])
```

With maybe None (and skipped fields) matching any POS component.
```dic.pos_matcher([('名詞', None)])``` will match any noun.

### PR DESCRIPTION
**Pull request by eiennohito**
fixes #177

### DIFF
--- a/python/src/pos_matcher.rs
+++ b/python/src/pos_matcher.rs
@@ -116,9 +116,8 @@ impl PyPosMatcher {
 #[pymethods]
 impl PyPosMatcher {
     pub fn __call__(&self, m: &PyMorpheme) -> bool {
-        let list = m.list();
-        let idx = m.index();
-        self.matcher.matches_id(list.get_word_info(idx).pos_id())
+        let pos_id = m.part_of_speech_id();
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: python/src/pos_matcher.rs [hunk 1]
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+}
+
+#[pymethods]
+impl PyPosMatcher {
+    pub fn __call__(&self, m: &PyMorpheme) -> bool {
+        let list = m.list();
+        let idx = m.index();
+        self.matcher.matches_id(list.get_word_info(idx).pos_id())
```
mh-northlander: How about using `m.part_of_speech_id()`?

mh-northlander (approved):

PR #185, opened 2021-11-30
implement analysis memory reuse via output parameters

### LINKED ISSUE #184: Pyhon API: ability to reduce allocations by reusing objects
`Tokenizer.analyze` and `Morpheme.split` will have new out parameter which will accept `MorphemeList`. It will be rewritten with new content. Passing the same MorphemeList twice will be diagnosed and an exception will be thrown.

eiennohito:
Fixes #184 

Implement out parameters for `Tokenizer.tokenize()` and `Morpheme.split()` Python API.

For the memory sharing to be actually useful, I had to refactor internal MorphemeList to allow multiple references to input data, while having distinct list of morphemes. Let's welcome `Arc<RefCell<X>>` in our codebase.
Python `MorphemeListWrapper` has also changed. As a side effect there is no copy in custom pretokenizers (win). Need to document semantics of everything, but will do that as a documentation pass.

### DIFF
--- a/python/tests/test_morpheme.py
+++ b/python/tests/test_morpheme.py
@@ -132,6 +132,22 @@ class TestTokenizer(unittest.TestCase):
         self.assertEqual(m[1].begin(), 2)
         self.assertEqual(len(m[0]), 2)
 
+    def test_morpheme_split_out(self):
+        ms = self.tokenizer_obj.tokenize('東京都', SplitMode.C)
+        self.assertEqual(1, ms.size())
+        self.assertEqual(ms[0].surface(), '東京都')
+
+        ms_a = ms[0].split(SplitMode.A, out=None)
+        self.assertEqual(2, ms_a.size())
+        self.assertEqual(ms_a[0].surface(), '東京')
+        self.assertEqual(ms_a[1].surface(), '都')
+
+        ms = self.tokenizer_obj.tokenize("京都東京都京都", SplitMode.C)
+        ms_b = ms[1].split(SplitMode.A, out=ms_a)
+        self.assertEqual(id(ms_a), id(ms_b))
+        self.assertEqual(ms_a[0].surface(), '東京')
+        self.assertEqual(ms_a[1].surface(), '都')
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/sudachi/tests/stateful_tokenizer.rs
+++ b/sudachi/tests/stateful_tokenizer.rs
@@ -58,19 +58,16 @@ fn get_word_id() {
 fn get_dictionary_id() {
     let mut tok = TestTokenizer::new_built(Mode::C);
     let ms = tok.tokenize("京都");
-    let ms: Vec<_> = ms.iter().collect();
     assert_eq!(1, ms.len());
-    assert_eq!(0, ms[0].dictionary_id());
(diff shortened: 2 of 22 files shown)


python/tests/test_morpheme.py
```diff
@@ -132,6 +132,22 @@ def test_normalize_katakana_half_full(self):
...
+        ms = self.tokenizer_obj.tokenize('東京都', SplitMode.C)
+        self.assertEqual(1, ms.size())
+        self.assertEqual(ms[0].surface(), '東京都')
+
+        ms_a = ms[0].split(SplitMode.A, out=None)
+        self.assertEqual(2, ms_a.size())
+        self.assertEqual(ms_a[0].surface(), '東京')
+        self.assertEqual(ms_a[1].surface(), '都')
+
+        ms = self.tokenizer_obj.tokenize("京都東京都京都", SplitMode.C)
```
mh-northlander, line 145: I want to check the second split with different word from the above one.

sudachi/tests/stateful_tokenizer.rs
```diff
@@ -130,18 +130,19 @@ fn split_middle() {
     let ms = tok.tokenize("京都東京都京都");
     assert_eq!(ms.len(), 3);
     let m = ms.get(1);
-    assert_eq!(m.surface(), "東京都");
+    assert_eq!(m.surface().deref().deref(), "東京都");
```
mh-northlander: double deref?

sudachi/tests/stateless_tokenizer.rs
```diff
@@ -35,34 +36,31 @@ fn get_word_id() {
     let ms = tok.tokenize("京都", Mode::C);
     let ms: Vec<_> = ms.iter().collect();
     assert_eq!(1, ms.len());
-    let pos = ms[0].part_of_speech().expect("failed to get pos");
+    let pos = ms[0].part_of_speech();
```
mh-northlander: We can now access nodes by `ms.get(0)` instead of `ms[0]`. (same for other test functions)

mh-northlander (changes requested):
`Arc<RefCell>` is not thread safe: https://doc.rust-lang.org/std/sync/struct.Arc.html#thread-safety
we may use `Mutex` or `RwLock` instead of `RefCell`.

python/tests/test_morpheme.py
eiennohito: Test binary dictionary has only 東京都 as a word with splits. I also wanted to check another word, but alas.
Moving all tests to using non-binary dictionary will fix this issue

eiennohito:
`Arc<RefCell<T>>` by itself is not thread-safe and that's OK. `StatefulTokenizer` and `MorphemeList` are not for sharing between threads. Additonally, for the Python bindings, all accesses to MorphemeList happen under GIL and are OK, for Rust the compiler will check invalid data sharing. 
After second thought, it is OK to make internals to be even `Rc<RefCell<T>>`.

sudachi/tests/stateless_tokenizer.rs
eiennohito: Tests need more cleanup, I agree.

eiennohito:
@mh-northlander comments should be addressed

mh-northlander (approved):

PR #188, opened 2021-12-02
use new capabilities of 0.6.1 in the Python command line

### LINKED ISSUE #170: Optimize python bindings
Profiling output of SudachiPy command line

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   440288   22.180    0.000   22.180    0.000 {method 'tokenize' of 'sudachipy.tokenizer.Tokenizer' objects}
        1   17.999   17.999   70.885   70.885 command_line.py:54(run)
 10811013   10.056    0.000   10.056    0.000 {method 'part_of_speech' of 'sudachipy.morpheme.Morpheme' objects}
 10811013    5.383    0.000    5.383    0.000 {method 'surface' of 'sudachipy.morpheme.Morpheme' objects}
 21622466    5.331    0.000    5.331    0.000 {method 'join' of 'str' objects}
 22062314    4.577    0.000    4.577    0.000 {method 'write' of '_io.TextIOWrapper' objects}
 10811013    4.522    0.000    4.522    0.000 {method 'normalized_form' of 'sudachipy.morpheme.Morpheme' objects}
 ```

1. Part of speech method returns a new list each time. Precalculate the whole possible list of POS tags in advance (as tuples) and return a precomputed tuple. #173 
2. Utilize partial WordInfo parsing to convert data into Python native types eagerly. #172

### PR DESCRIPTION
**Pull request by eiennohito**
fixes #170

### DIFF
--- a/python/py_src/sudachipy/command_line.py
+++ b/python/py_src/sudachipy/command_line.py
@@ -19,8 +19,8 @@ import os
 import sys
 from pathlib import Path
 
-from . import __version__
 from . import Dictionary, SplitMode
+from . import __version__
 from . import sudachipy
 
 
@@ -55,9 +55,10 @@ def run(tokenizer, input_, output, print_all, morphs, is_stdout):
     for line in input_:
         line = line.rstrip('\n')
         # Note: Current version of the tokenizer ignores logger
+        # use out parameter to reuse MorphemeList allocated memory
         mlist = tokenizer.tokenize("")
         for m in tokenizer.tokenize(line, out=mlist):
-             list_info = [
+            list_info = [
                 m.surface(),
                 ",".join(morphs[m.part_of_speech_id()]),
                 m.normalized_form()]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: python/py_src/sudachipy/command_line.py [hunk 1]
```diff
@@ -51,26 +51,37 @@ def _set_default_subparser(self, name, args=None):
...
+def run(tokenizer, input_, output, print_all, morphs, is_stdout):
     for line in input_:
         line = line.rstrip('\n')
         # Note: Current version of the tokenizer ignores logger
-        for m in tokenizer.tokenize(line, mode, logger if enable_dump else None):
-            list_info = [
-                m.surface(),
-                ",".join(m.part_of_speech()),
-                m.normalized_form()]
+        mlist = tokenizer.tokenize("")
```
mh-northlander: It would be nice to have a comment about reuse technique of morpheme list

mh-northlander (approved):

python/py_src/sudachipy/command_line.py
mh-northlander: Can we move this to outside of the first for-loop?

eiennohito: Super nice catch, I completely missed this one

PR #211, opened 2022-06-15
Regex-based oov handler

eiennohito:
Add a regex-based OOV handler in sync with the [Java version](https://github.com/WorksApplications/Sudachi/blob/develop/src/main/java/com/worksap/nlp/sudachi/RegexOovProvider.java)

Differences:

* Rust Regex crate does not support backreferences and loookahead (but I want to forbid them in Java as well)
* Rust OOV handler accepts `oovPOS` as an alias of `pos` setting key, which Java does not.
* Rust OOV handler better handles non-start of string matches.
* Rust `maxLength` is in unicode codepoints, Java `maxLength` is in utf-8 bytes, but I will port Java version to codepoints later (infra does not support it cleanly yet).

Rest should be identical


sudachi/src/dic/grammar.rs
```diff
@@ -109,15 +110,45 @@ impl<'a> Grammar<'a> {
...
+    pub fn get_part_of_speech_id<S>(&self, pos1: &[S]) -> Option<u16>
+    where
+        S: AsRef<str>,
+    {
+        if pos1.len() != POS_DEPTH {
+            return None;
+        }
         for (i, pos2) in self.pos_list.iter().enumerate() {
-            if pos1.len() == pos2.len() && pos1.iter().zip(pos2).all(|(a, b)| a == b) {
+            if pos1.len() == pos2.len() && pos1.iter().zip(pos2).all(|(a, b)| a.as_ref() == b) {
```
mh-northlander: We can omit the length check (`pos1.len() == pos2.len()`), thanks to the check added above

mh-northlander (approved):
LGTM

PR #219, opened 2022-07-22
add boundary mode parameter to regex oov handler

eiennohito:
port of https://github.com/WorksApplications/Sudachi/pull/193


mh-northlander (approved):
I think it's nice to put common documents in Java side, but it's kinder to add some note about it in README

eiennohito:
I agree that documentation explorability(?) needs some work

PR #221, opened 2022-10-05
add Dictionary.lookup method

eiennohito:
it fetches entries from the dictionary without doing analysis

### DIFF
--- a/python/src/dictionary.rs
+++ b/python/src/dictionary.rs
@@ -32,6 +32,7 @@ use sudachi::plugin::input_text::InputTextPlugin;
 use sudachi::plugin::oov::OovProviderPlugin;
 use sudachi::plugin::path_rewrite::PathRewritePlugin;
 
+use crate::morpheme::PyMorphemeListWrapper;
 use crate::pos_matcher::PyPosMatcher;
 use crate::pretokenizer::PyPretokenizer;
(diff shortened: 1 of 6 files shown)


python/src/dictionary.rs
```diff
@@ -271,6 +272,37 @@ impl PyDictionary {
             .call1(PyTuple::new(py, [internal_cell]))
     }
 
+    #[pyo3(text_signature = "($self, surface, list = None) -> sudachipy.MorphemeList")]
```
mh-northlander: 3rd arg `list` seems not renamed (provably `out`).

mh-northlander (changes requested):
Error occurs when I call `end` of resulting Morpheme:

```
> dict = sudachipy.Dictionary()
> mrps = dict.lookup("東京都")
> mrps[0].end()
thread '<unnamed>' panicked at 'index out of bounds: the len is 4 but the index is 9', sudachi-lib/src/input_text/buffer/mod.rs:332:24
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pyo3_runtime.PanicException: index out of bounds: the len is 4 but the index is 9
```

Maybe the byte/char length conversion is the cause.

eiennohito:
@mh-northlander thanks for the catch, fixed logic and added tests, both in Python and Rust for the logic

mh-northlander (approved):

PR #240, opened 2024-01-22
fetch latest dictionary

mh-northlander:
update `fetch_dictionary.sh`

- download latest Sudachi dictionary
- take version and dict_type as arguments (default: latest / core)

### DIFF
--- a/.github/workflows/build-python-wheels.yml
+++ b/.github/workflows/build-python-wheels.yml
@@ -41,7 +41,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Download dictionary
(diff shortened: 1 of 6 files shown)


eiennohito:
No, the downloaded dictionary (at least in CI) should be fixed because the CI tests the tokenization against fixed results.

mh-northlander:
> No, the downloaded dictionary (at least in CI) should be fixed because the CI tests the tokenization against fixed results.

I missed that point.

We can do either of
- make the default dictionary-version of `fetch_dictionary.sh` fixed
- let CI scripts call `fecth_dictionary.sh` with fixed version

I prefer the latter since it's more explicit.

PR #245, opened 2024-02-29
Allow string literals as segmentation modes

### LINKED ISSUE #243: Update SplitMode implementation
The issue around enum on PyO3 side seems solved: https://github.com/PyO3/pyo3/issues/834.
https://github.com/WorksApplications/sudachi.rs/blob/d09dd0079b48cd307626b7a4a350deddd52ea132/python/src/tokenizer.rs#L38-L39

Check if we can use it for the PySplitMode.

eiennohito:
Fixes #243 

In addition to `SplitMode` objects SudachiPy will now accept A, B, C string literals as well (in all places).
Split mode also gets a constructor that accepts string literal in addition to class members

Bonus fix: tokenization now releases GIL

### DIFF
--- a/python/py_src/sudachipy/sudachipy.pyi
+++ b/python/py_src/sudachipy/sudachipy.pyi
@@ -1,5 +1,5 @@
 from typing import ClassVar, Iterator, List, Tuple, Union, Callable, Iterable, Optional, Literal, Set
-from sudachipy.config import Config
+from .config import Config
 
 POS = Tuple[str, str, str, str, str, str]
 # POS element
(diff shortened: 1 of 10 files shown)


python/py_src/sudachipy/sudachipy.pyi
```diff
@@ -65,7 +70,7 @@ class Dictionary:
         ...
 
     def create(self,
-               mode: SplitMode = SplitMode.C,
+               mode: Union[SplitMode, Literal["A"], Literal["B"], Literal["C"]] = SplitMode.C,
```
mh-northlander: Let's use `Union[SplitMode, Literal["A", "B", "C"]]` as l.199.

python/src/dictionary.rs
mh-northlander, line 1: `Dictionary.pre_tokenizer` also take split mode as an argument (l.292).

eiennohito:
@mh-northlander fixed!

PR #255, opened 2024-06-03
Aggregate package information into root Cargo.toml

### LINKED ISSUE #163: Find out and document minimum-supported rust version and add it to test matrix, at least for Linux

mh-northlander:
Use [workspace package table](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) to write package information.

Also add minimum-support rust version to resolve #163.

### DIFF
--- a/update_version.sh
+++ b/update_version.sh
@@ -2,12 +2,12 @@
 
 set -eu
 
-if [ $# -lt 2 ] && [ "$1" != "show" ] ; then
+if [ $# -lt 1 ] || ( [ $# -lt 2 ] && [ "$1" != "show" ] ) ; then
     echo "Provide 2 arguments [from] and [to] to update version, or 'show' to print current one."
     exit 1
 fi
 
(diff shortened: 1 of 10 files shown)


update_version.sh
```diff
@@ -2,12 +2,12 @@
 
 set -eu
 
-if [ $# -lt 2 ] && [ "$1" != "show" ] ; then
+if [ $# -lt 1 ] || ( [ $# -lt 2 ] && [ "$1" != "show" ] ) ; then
```
kazuma-t, line 5: Since the version descriptions have been consolidated into one file, isn't this script unnecessary?

update_version.sh
mh-northlander, line 1: We keep this script for now, because we still need to embed a version string to several files in python side.

kazuma-t: I understand.

kazuma-t (approved):

Issue #218, opened 2022-07-18
Can the default resource files be embedded on compile time?


yutannihilation:
Hi, thanks for developing this great library. I played a bit with R ([my repo](https://github.com/yutannihilation/fledgingr)), but I had to struggle with this error when I tried to make a static binary for users that don't have Rust installed. It seems this error is because `Config::new()` cannot find the files under `resources` directory.

```
Io(Os { code: 3, kind: NotFound, message: "指定されたパスが見つかりません。" })
```

As the files are not very huge, is it possible to embed them somehow, e.g. using `include_str` macro, on compile time?

eiennohito:
Config framework needs to have a complete overhaul. Everything which is present now is a placeholder.

I agree that defaults probably should be embedded in the binary/library as resources.

PR #262, opened 2024-06-13
Add option for embedded config and fallback resources

Kuuuube:
Partial fix for #218

Allows static binaries to function properly in rust by falling back on default config and resources. Missing support for user defined embedded files but I think sudachi.rs should have at least basic support for static binaries.

Basic example code creating a tokenizer using embedded defaults (dictionary must still be added separately):

```rust
let embedded_dictionary = include_bytes!("./system_full.dic");
let sudachi_dic_data = SudachiDicData::new(Storage::Borrowed(embedded_dictionary));
let config = Config::new_embedded().unwrap();
let japanese_dictionary = JapaneseDictionary::from_cfg_storage_with_embedded_chardef(&config, sudachi_dic_data).unwrap();

let tokenizer = StatelessTokenizer::new(&japanese_dictionary);
```

### DIFF
--- a/sudachi/src/config.rs
+++ b/sudachi/src/config.rs
@@ -29,6 +29,7 @@ use thiserror::Error;
 
 const DEFAULT_RESOURCE_DIR: &str = "resources";
 const DEFAULT_SETTING_FILE: &str = "sudachi.json";
+const DEFAULT_SETTING_BYTES: &[u8] = include_bytes!("../../resources/sudachi.json");
 const DEFAULT_CHAR_DEF_FILE: &str = "char.def";
 
 /// Sudachi Error
@@ -339,6 +340,13 @@ impl Config {
             None => raw_config,
             Some(p) => raw_config.system_dict(p),
         };
+
+        Ok(raw_config.build())
+    }
+
+    pub fn new_embedded() -> Result<Self, ConfigError> {
+        // prioritize arg (cli option) > default
+        let raw_config = ConfigBuilder::from_bytes(DEFAULT_SETTING_BYTES)?;
 
         Ok(raw_config.build())
     }
--- a/sudachi/src/config.rs
+++ b/sudachi/src/config.rs
@@ -345,7 +345,6 @@ impl Config {
     }
 
     pub fn new_embedded() -> Result<Self, ConfigError> {
-        // prioritize arg (cli option) > default
         let raw_config = ConfigBuilder::from_bytes(DEFAULT_SETTING_BYTES)?;
 
         Ok(raw_config.build())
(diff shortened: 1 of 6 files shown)


sudachi/src/config.rs
```diff
@@ -343,6 +344,13 @@ impl Config {
         Ok(raw_config.build())
     }
 
+    pub fn new_embedded() -> Result<Self, ConfigError> {
+        // prioritize arg (cli option) > default
```
mh-northlander: this comment should be removed (this method does not take args from cli)

sudachi/src/dic/mod.rs
```diff
@@ -71,6 +71,25 @@ impl<'a> LoadedDictionary<'a> {
         })
     }
 
+    /// Creates a system dictionary from bytes, and load embedded default character category
+    pub fn from_system_dictionary_embedded(
+        dictionary_bytes: &'a [u8],
+    ) -> SudachiResult<LoadedDictionary<'a>> {
+        let system_dict = DictionaryLoader::read_system_dictionary(dictionary_bytes)?;
+
+        let character_category = CharacterCategory::from_bytes(include_bytes!("../../../resources/char.def"))?;
```
mh-northlander: Let's add const `DEFAULT_CHAR_DEF_BYTES` for this include bytes.

sudachi/src/dic/mod.rs
```diff
@@ -71,6 +71,25 @@ impl<'a> LoadedDictionary<'a> {
         })
     }
 
+    /// Creates a system dictionary from bytes, and load embedded default character category
+    pub fn from_system_dictionary_embedded(
```
mh-northlander, line 87: This method shares main logic with `from_system_dictionary`.
Let's add following method to split shared part.

```
pub fn from_system_dictionary_and_chardef(
        dictionary_bytes: &'a [u8],
        character_category: CharacterCategory,
)
```

sudachi/src/dic/dictionary.rs
```diff
@@ -115,6 +115,43 @@ impl JapaneseDictionary {
         Ok(dic)
     }
 
+    /// Creates a dictionary from the default embedded configuration and storage
+    pub fn from_embedded_storage(
```
mh-northlander: I want to give more explicit name for this method like:
```
/// Creates a dictionary from the specified configuration and storage, with embedded character definition
pub fn from_cfg_storage_with_embedded_chardef(...
```

mh-northlander (changes requested):
Ideally the embedded resource files should be used as a system default (replacing current default-resource-directory based configuration).
But as a first step we can have both.

I accept this PR with some miner changes.

Kuuuube:
Thank you for the review! Everything should be resolved.

mh-northlander (approved):

mh-northlander:
@Kuuuube  please apply `cargo fmt`.

Kuuuube:
Done

Issue #220, opened 2022-08-22
aarch64 Linux Wheels


polm:
I noticed you don't have wheels for aarch64 for Linux. Do you have any plans to release such wheels? 

Github Actions supports this architecture, so it looks like you can probably just make a small change to the current wheel building actions.

eiennohito:
Since the last version, we do universal2 builds for macOS which work both on aarch64 and x64 at the same time, which is better in my opinion. The bad thing is Github Actions does not support managed runners yet, only self-hosted ones and we neither have infrastructure nor manpower to host the runner ourselves. So while there are build, they are not tested on aarch64 at all and may stop working at any time. I run tests on my M1 Macbook from time to time, but until there is a hosted option for Github Actions runner, it will not move futher.

cf:
1. https://github.com/github/roadmap/issues/528
2. https://github.com/actions/runner-images/issues/2187

polm:
Thanks for the clarification, but I am asking specifically about Linux aarch64, not OSX. I know the managed OSX runners on Github Actions still aren't fully usable with Python, but with Linux they should work - I build aarch64 wheels for mecab-python3 and fugashi. For example [here](https://github.com/polm/fugashi/blob/cbd2255a5d711698d7f4d55ae59c7d11c9293153/.github/workflows/manylinux1.yml#L62) is the action for aarch64 wheels for fugashi.

I don't think Linux aarch64 is that common, but it does include Raspberry Pi users, and potentially people running Linux (or Docker) on M1 machines.

eiennohito:
Ah, Linux aarch64. I completely missed Linux in your first post, apologies. 

Pull request is welcome! Current linux build does PGO which is good for binary builds, but it probably should work on aarch64 as well with minor modifications.

Rhtyme:
In my case, I just installed rust :

`curl --proto '=https' --tlsv1.3 https://sh.rustup.rs -sSf | sh`
and then did :

`source $HOME/.cargo/env`
by following [this tutorial](https://www.digitalocean.com/community/tutorials/install-rust-on-ubuntu-linux).

Then I just did:

`pip install --upgrade 'sudachipy>=0.6.8'`

as it is instructed, and installed successfully!

tsukumijima:
@eiennohito @mh-northlander Sudachipy 0.6.9, released this month, has the following error in Linux arm64 environment, and wheel cannot be built itself.

```bash
Defaulting to user installation because normal site-packages is not writeable
[...]

PR #272, opened 2024-09-16
Add wheel build for the aarch64 architecture in Github actions

lafeuil:
Add in the `Build Linux Python Wheels (+PGO)` job of the `Python Packages` Github actions workflow, a new `arch` matrix variable to build the wheel on the `aarch64` architecture.

The wheel already builds fine on the `aarch64` architecture in local.

This PR fixes the #220 issue.

I didn't test this new configuration in the Github workflow because I don't know how to test it.

### DIFF
--- a/.github/workflows/build-python-wheels.yml
+++ b/.github/workflows/build-python-wheels.yml
@@ -38,6 +38,9 @@ jobs:
   build-linux-wheels:
     name: Build Linux Python Wheels (+PGO)
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [none, aarch64]
     steps:
       - uses: actions/checkout@v4
 
@@ -56,6 +59,7 @@ jobs:
 
       - uses: eiennohito/gha-manylinux-build@master
         with:
+          arch: ${{ matrix.arch }}
           script: python/build-wheels-manylinux-pgo.sh
 
       - uses: actions/upload-artifact@v4


.github/workflows/build-python-wheels.yml
```diff
@@ -56,6 +59,7 @@ jobs:
 
       - uses: eiennohito/gha-manylinux-build@master
         with:
+          arch: ${{ matrix.arch }}
           script: python/build-wheels-manylinux-pgo.sh
 
       - uses: actions/upload-artifact@v4
```
mh-northlander, line 65: [`upload-artifact@v4` does not allow uploading multiple times with same name](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes).
Add matrix parameter to `name` input (it must match `artifact-*` pattern in l.120).

Also, we merge contents of each artifacts with `download-artifact@v4` in `upload-to-pypi` job.
The names of built files should be changed not to overwrite other architecture's.

.github/workflows/build-python-wheels.yml
```diff
@@ -56,6 +59,7 @@ jobs:
 
       - uses: eiennohito/gha-manylinux-build@master
```
mh-northlander, line 60: [`eiennohito/gha-manylinux-build`](https://github.com/eiennohito/gha-manylinux-build) does not have `arch` input.
I guess we need a different method to build aarch64 wheel.

mh-northlander (changes requested):

.github/workflows/build-python-wheels.yml
eiennohito: I can update the build definition if somebody makes a PR for it. Or you can clone the repo to your user and use the clone definiton.

PR #281, opened 2024-11-19
Pre v0.6.9

mh-northlander:


mh-northlander:
The test failed due to the lag in TestPyPI (the post step could not download the uploaded package).
I skip the test for now.

PR #282, opened 2024-12-02
fix dictionary.__repr__ for no user-dict case

mh-northlander:
current `__repr__` misses close "]" when the user dict list is empty:

```python
dict = sudachipy.Dictionary()
dict.__repr__()
# '<SudachiDictionary(system=/path/to/system.dic, user=[)>'
```

### DIFF
--- a/python/src/dictionary.rs
+++ b/python/src/dictionary.rs
@@ -434,10 +434,10 @@ fn config_repr(cfg: &Config) -> Result<String, std::fmt::Error> {
     match cfg.resolved_user_dicts() {
         Ok(dicts) => {
             for (i, dic) in dicts.iter().enumerate() {
-                write!(result, "{}", dic.display())?;
-                if i + 1 != dicts.len() {
+                if i != 0 {
                     write!(result, ", ")?;
                 }
+                write!(result, "{}", dic.display())?;
             }
         }
         Err(e) => {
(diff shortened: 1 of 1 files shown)


python/src/dictionary.rs
```diff
@@ -435,9 +435,7 @@ fn config_repr(cfg: &Config) -> Result<String, std::fmt::Error> {
         Ok(dicts) => {
             for (i, dic) in dicts.iter().enumerate() {
                 write!(result, "{}", dic.display())?;
-                if i + 1 == dicts.len() {
-                    write!(result, "]")?;
-                } else {
+                if i + 1 != dicts.len() {
```
kazuma-t: Wouldn't `i < dict.len()` be better?

kazuma-t (changes requested):

python/src/dictionary.rs
kazuma-t: Sorry, `i < dict.len() - 1`. Both are equally complicated, so you probably don't need to fix it.

mh-northlander:
`Vec<T>` does not have the display trait and we need to write.
https://doc.rust-jp.rs/rust-by-example-ja/hello/print/print_display.html

I chose to branch at index=0.

kazuma-t (approved):

PR #295, opened 2025-01-08
introduce cibuildwheel to build workflow

mh-northlander:
Introduce [cibuildwheel](https://github.com/pypa/cibuildwheel) to simplify the build workflow.


mh-northlander:
note:
arm64 emulation on x86 ubuntu using QEMU seems too slow for us (compilation takes hours and canceled).
[pypa/cibuildwheel](https://github.com/pypa/cibuildwheel) seems useful but internally uses same emulation.

arm64 runner as a larger runners (pay-as-you-go) are available: [blog post](https://github.blog/changelog/2024-09-03-github-actions-arm64-linux-and-windows-runners-are-now-generally-available/)
arm64 linux runner is planed to be available in early 2025 (ref: https://github.com/orgs/community/discussions/19197).

kazuma-t (approved):

PR #311, opened 2026-03-04
add Cargo.lock to python sdist

### LINKED ISSUE #309: Lock dependency versions
Hi

We're using `sudachi` for Japanese language tokenization in our product, but recently had some build problems in CI. After doing a bit of digging around, I think part of the problem is that `sudachi.rs` doesn't pin its Cargo dependencies and so they can change on every build. 

I'm not a Rustacean, so I might be wrong, but my understanding is that without a `Cargo.lock` file, `cargo` will simply download the latest semver compatible version of the library. Unfortunately, [indexmap had a recent change that required an updated Rust compiler without releasing a new major version](https://github.com/indexmap-rs/indexmap/issues/429), so these changing dependencies broke our build.

Please could you look at adding pinning dependencies, both on the Python side and Rust side, allowing builds of `sudachi.rs` to be reproducible and avoiding such issues.

Thank you for your hard work maintaining this package!

hayashi-mas-wap:
fix #309

- update Cargo.lock
- add Cargo.lock to python sdist (copied from root)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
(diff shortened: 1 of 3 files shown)


yanagisawa-yu-wap (approved):
LGTM 👍

PR #316, opened 2026-03-19
implement dictionary format v1

hayashi-mas-wap:
Implement dictionary format V1 compatible with [Java v0.8](https://github.com/WorksApplications/Sudachi/tree/develop-v0.8).

Since it is hard to compare dictionary binaries itself, we check followings (let `java.dic` be a binary dic built with Java version):
- basics
  - [x] we can build binary from v0 format csv (`v0.dic`)
  - [x] we can build binary from v1 format csv (`v1.dic`)
  - [x] we can run analysis with built dictionary
  - [x] we can dump lexicon csv from built dictionary  
- analysis compatibility 
  - [x] All of analysis result using `v0.dic`, `v1.dic`, `java.dic` should be same
  - [x] Check the same thing using java tokenizer
  - ~java/rust analysis results should be same~
- dump compatibility
  - [x] All of lexicon dump of `v0.dic`, `v1.dic`, `java.dic` should be same
  - [x] Check the same thing using java `DictionaryPrinter`
  - [x] java/rust dump should be same
- dump + rebuild
  - [x] we can build binary from dumped lexicon (`dump.dic`)
  - [x] analysis result with `dump.dic` should be same
  - [x] dump result from `dump.dic` should be same


hayashi-mas-wap:
There are some analysis result difference, but they do not relate to the dictionary format.
Separate them as another issue: #317

takaoka-k-wap (approved):

Issue #317, opened 2026-04-01
Sync Java/rust analysis behavior


hayashi-mas-wap:
There are some difference in the analysis of Java/rust version.
We want to sync those.

- [x] sample text: "大自然ー空や海、湖、森"
  - `char.def` is extended in rust. They are used in `src/input_text/buffer.rs::build` to calc `can_bow`, however, those are "no-oov-bow" and probably should not handled at there.
- [x] sample text: "「禰々」"
- [x] sample text: "いつ見てもオニカサゴの釣果が安定している"
  - pos of katakana OOV word become "感動詞,一般,*,*,*,*" in Java ("名詞,普通名詞,一般,*,*,*" in rust).
  - This seems the problem in Java side.

PR #325, opened 2026-04-10
Fix usage of NOOOVBOW char category

hayashi-mas-wap:
relates to #317.

This will fix the problem of "大自然ー空や海、湖、森" (「大自然」will be a single morpheme).

- NOOOVBOW should be used only for OOV: stop using it for `input_buffer.can_bow`, and use it in `build_lattice`.
- Introduce NOOOVEOW category, and change NOOOVBOW2 to (NOOOVBOW + NOOOVEOW)

Latter (and char.def update) should be ported to Java.


hayashi-mas-wap:
Maybe the name `NOOOVEOW` is a bit misleading as it CAN be a end of OOV word. It just restricts the next character to be a OOVBOW.

`NOOOVAFTER` or `OOVJOINNEXT`?

takaoka-k-wap (approved):

PR #330, opened 2026-05-11
Fix post-#263 clippy warnings in sudachi crate

gulldan:
Partial fix for #40.

This fixes default-level clippy/rustc warnings introduced after #263 in the `sudachi` crate only.
No API or behavior changes are intended.

Fixed:
- `mismatched_lifetime_syntaxes`: make elided lifetimes explicit on return types
- `needless_as_bytes`
- `needless_return`
- `unnecessary_lazy_evaluations`
- `unnecessary_sort_by`
- `missing_transmute_annotations`
- `needless_lifetimes`
- `empty_line_after_doc_comments`

Intentionally left out:
- `result_large_err`
- dead-code warnings
- `&Vec` / `&String` signature changes
- `StringNumber::to_string` rename
- `sudachi-fuzz` `unexpected_cfgs(fuzzing)`
- `python/` transmute annotations

Tested:
- `cargo fmt -- --check --files-with-diff`
- `cargo test --verbose`

### DIFF
--- a/sudachi/src/analysis/mlist.rs
+++ b/sudachi/src/analysis/mlist.rs
@@ -135,16 +135,16 @@ impl<T: DictionaryAccess> MorphemeList<T> {
         self.nodes.data.is_empty()
     }
 
-    pub fn get(&self, idx: usize) -> Morpheme<T> {
-        return Morpheme::for_list(self, idx);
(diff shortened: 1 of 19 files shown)


hayashi-mas-wap:
Thank you for your contribution, however, as we have large diffs in `develop-v0.7`, I rather want to skip this fix in `develop`.

PR #333, opened 2026-05-11
Add morpheme form entry accessors

### LINKED ISSUE #322: Get a morpheme for the dictionary-form / the normalized-form of a morpheme
relates to https://github.com/WorksApplications/Sudachi/issues/265

gulldan:
## Summary
- implement morpheme form entry accessors in Rust core, not only in Python bindings
- keep the existing list-backed Rust `Morpheme` type name and add `MorphemeView` as the shared accessor trait
- add `SingleMorpheme` and `MorphemeRef` for exact-`WordId` standalone dictionary entries, including standalone split support and `user_data()` access
- guard invalid, OOV, and special `WordId`s in public exact-entry materialization APIs
- delegate Python `dictionary_form_morpheme()` / `normalized_form_morpheme()` to Rust core and back Python `Morpheme` / `MorphemeList` with list-backed or `SingleMorpheme` variants while preserving the public Python API
- materialize form morphemes with the source `InfoSubset + HEADWORD`, while loading only the required form-reference field internally
- align standalone split behavior with Java empty / single / multi split semantics, including parent span preservation for single replacement splits and lazy split metadata loading
- preserve standalone split behavior for Python form morphemes and add Rust/Python regression coverage
- add changelog entries for the Rust and Python API changes

Fixes #322.

## Validation
- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo check -p sudachi`
- `cargo check -p sudachipy`
- `cargo test -p sudachi`
- `cargo test -p sudachipy`
- `cargo test --workspace`
- `cd python && .env/bin/python -m unittest tests.test_morpheme -v`
- `cd python && bash build_and_test.sh`
- Cross-runtime compatibility check against Java `develop-v0.8`, Rust, and Python with a rebuilt full SudachiDict `20260428` dictionary

### DIFF
--- a/python/src/morpheme.rs
+++ b/python/src/morpheme.rs
@@ -331,7 +331,7 @@ impl PyMorpheme {
 
         let mut form_list = MorphemeList::empty(dict);
         errors::wrap_ctx(
-            form_list.lookup_word_id(form_word_id, InfoSubset::all()),
+            form_list.reset_with_word_id(form_word_id, InfoSubset::all()),
             context,
         )?;
 
@@ -409,7 +409,10 @@ impl PyMorpheme {
 
     /// Returns the morpheme corresponding to this morpheme's dictionary form.
     ///
-    /// If this morpheme is out-of-vocabulary, returns a morpheme equivalent to ``self``.
+    /// For out-of-vocabulary morphemes, invalid references, and references to
+    /// the same dictionary entry, returns a morpheme equivalent to ``self``.
+    /// For a distinct referenced entry, returns a standalone morpheme whose
+    /// begin/end offsets are ``0..len(surface)``.
     #[pyo3(text_signature = "(self, /) -> Morpheme")]
     fn dictionary_form_morpheme<'py>(&'py self, py: Python<'py>) -> PyResult<PyMorpheme> {
         self.form_morpheme(
@@ -428,7 +431,10 @@ impl PyMorpheme {
 
     /// Returns the morpheme corresponding to this morpheme's normalized form.
     ///
-    /// If this morpheme is out-of-vocabulary, returns a morpheme equivalent to ``self``.
+    /// For out-of-vocabulary morphemes, invalid references, and references to
+    /// the same dictionary entry, returns a morpheme equivalent to ``self``.
+    /// For a distinct referenced entry, returns a standalone morpheme whose
+    /// begin/end offsets are ``0..len(surface)``.
     #[pyo3(text_signature = "(self, /) -> Morpheme")]
     fn normalized_form_morpheme<'py>(&'py self, py: Python<'py>) -> PyResult<PyMorpheme> {
         self.form_morpheme(
--- a/python/src/morpheme.rs
+++ b/python/src/morpheme.rs
@@ -318,7 +318,9 @@ impl PyMorpheme {
                     list.projection.clone(),
                     morpheme.word_id(),
                 ),
-                _ => unreachable!("unhandled Rust morpheme reference variant"),
+                _ => {
+                    return errors::wrap(Err("unsupported Rust morpheme reference variant"));
+                }
             }
         };
 
--- a/python/src/morpheme.rs
+++ b/python/src/morpheme.rs
@@ -280,13 +280,20 @@ impl PyMorpheme {
         )
     }
 
+    fn self_equivalent(&self, py: Python<'_>) -> PyMorpheme {
+        PyMorpheme {
+            list: self.list.clone_ref(py),
+            index: self.index,
+        }
+    }
+
     fn form_morpheme<'py, F>(
         &'py self,
         py: Python<'py>,
         subset: InfoSubset,
         form_word_id: F,
         context: &str,
-    ) -> PyResult<Option<PyMorpheme>>
+    ) -> PyResult<PyMorpheme>
     where
         F: FnOnce(&WordInfoData) -> WordId,
     {
@@ -296,7 +303,7 @@ impl PyMorpheme {
             let word_id = internal.get(self.index).word_id();
 
             if word_id.is_oov() || word_id.is_special() {
-                return Ok(None);
+                return Ok(self.self_equivalent(py));
             }
 
             let word_info = errors::wrap_ctx(
@@ -307,8 +314,12 @@ impl PyMorpheme {
                 context,
             )?;
             let form_word_id = form_word_id(word_info.borrow_data());
-            if form_word_id.is_oov() || form_word_id.is_special() {
-                return Ok(None);
+            if form_word_id == WordId::INVALID
+                || form_word_id == word_id
+                || form_word_id.is_oov()
+                || form_word_id.is_special()
+            {
+                return Ok(self.self_equivalent(py));
             }
 
             (
@@ -328,10 +339,10 @@ impl PyMorpheme {
             py,
             PyMorphemeListWrapper::from_components(form_list, projection),
         )?;
-        Ok(Some(PyMorpheme {
+        Ok(PyMorpheme {
             list: py_list,
             index: 0,
-        }))
+        })
     }
 }
 
@@ -398,9 +409,9 @@ impl PyMorpheme {
 
     /// Returns the morpheme corresponding to this morpheme's dictionary form.
     ///
-    /// If this morpheme is out-of-vocabulary, returns ``None``.
-    #[pyo3(text_signature = "(self, /) -> Morpheme | None")]
-    fn dictionary_form_morpheme<'py>(&'py self, py: Python<'py>) -> PyResult<Option<PyMorpheme>> {
+    /// If this morpheme is out-of-vocabulary, returns a morpheme equivalent to ``self``.
+    #[pyo3(text_signature = "(self, /) -> Morpheme")]
+    fn dictionary_form_morpheme<'py>(&'py self, py: Python<'py>) -> PyResult<PyMorpheme> {
         self.form_morpheme(
             py,
             InfoSubset::DICTIONARY_FORM,
@@ -417,9 +428,9 @@ impl PyMorpheme {
 
     /// Returns the morpheme corresponding to this morpheme's normalized form.
     ///
-    /// If this morpheme is out-of-vocabulary, returns ``None``.
-    #[pyo3(text_signature = "(self, /) -> Morpheme | None")]
-    fn normalized_form_morpheme<'py>(&'py self, py: Python<'py>) -> PyResult<Option<PyMorpheme>> {
+    /// If this morpheme is out-of-vocabulary, returns a morpheme equivalent to ``self``.
+    #[pyo3(text_signature = "(self, /) -> Morpheme")]
+    fn normalized_form_morpheme<'py>(&'py self, py: Python<'py>) -> PyResult<PyMorpheme> {
         self.form_morpheme(
             py,
             InfoSubset::NORMALIZED_FORM,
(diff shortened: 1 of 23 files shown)


hayashi-mas-wap (changes requested):
This feature should also be implemented in Rust side.
And, following Java implementation, we want to have `SingleMorphemeImpl` like struct and `Morpheme` like interface.

gulldan:
Thanks, I see the gap now. The current PR exposes the feature only through the Python binding and only adds a low-level Rust helper.

I’ll move the referenced-form resolution into the Rust core by adding a standalone morpheme type analogous to Java’s SingleMorphemeImpl and a common morpheme abstraction for list-backed and standalone morphemes. Then Rust users will get dictionary_form_morpheme()/normalized_form_morpheme() directly, and the Python binding will delegate to that core behavior instead of carrying separate resolver logic.

gulldan:
I pushed a follow-up commit (`62f2c735`) that tightens the Rust core implementation and the Python bridge after the previous review.

What changed in this follow-up:
- Kept the existing public list-backed Rust `Morpheme` type name, and introduced `MorphemeView` as the shared accessor trait instead of replacing the old `Morpheme` type with a trait.
- Added `user_data()` to the Rust morpheme interface/wrappers.
- Marked `MorphemeRef` as `#[non_exhaustive]` and sealed the shared trait implementation surface.
- Added guards for invalid, OOV, and special `WordId`s in exact-entry materialization paths, returning `SudachiError::InvalidWordId` instead of allowing lexicon indexing panics.
- Added `SingleMorpheme::split()` for standalone morphemes, matching the Java-style standalone form-entry behavior.
- Added direct Rust coverage for `SingleMorpheme::from_word_id()` homographs, invalid word IDs, standalone splits, and `user_data()` access.
- Added Python stub entries and a regression test for splitting form morphemes returned from Rust core.

One issue found during full-dictionary compatibility testing:
- Python form morphemes returned from a Rust `SingleMorpheme` were being rewrapped as a one-entry `MorphemeList`. That worked for simple accessors, but could break split offsets for standalone referenced entries. With a full dictionary case such as a normalized form whose split surfaces are not bounded by the normalized surface string, this path could panic in the Rust input buffer. The follow-up keeps the Python representation unchanged, but records the standalone `WordId` and routes `split()` through Rust `SingleMorpheme::split()`.

Cross-runtime check:
- Java: `WorksApplications/Sudachi` `develop-v0.8` was used because the Maven `0.7.5` release does not yet expose `dictionaryFormMorpheme()` / `normalizedFormMorpheme()`.
- Dictionary: rebuilt a compatible full dictionary from SudachiDict `20260428` raw CSVs because the prebuilt latest full dictionary is legacy-format for the current Rust/Python/Java development code.
- Corpus: 2,000 Japanese Wikipedia title lines.
- Result: Java/Rust/Python produced byte-for-byte identical TSV output for the accessor/split probe: 10,573 morpheme rows, SHA-256 `2f058e52dc4d066fea45606c9a9f6a36881b59c53b6f45770488e7fba2f577bc`.

Rough local benchmark on 99,999 Wikipedia title lines, 504,567 morphemes, with tokenization + form accessors + dictionary/normalized form morphemes + A/B splits:

| Impl | Elapsed | Morphemes/sec |
| --- | ---: | ---: |
| Java develop-v0.8 | 1.400s | 360,438 |
| Rust local | 1.550s | 325,432 |
| Python local editable/debug | 14.654s | 34,432 |

Validation run locally:
- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo check -p sudachi`
- `cargo check -p sudachipy`
- `cargo test -p sudachi`
- `cargo test -p sudachipy`
- `cd python && .env/bin/python -m unittest tests.test_morpheme -v`
- `cd python && bash build_and_test.sh`

python/tests/test_morpheme.py
```diff
@@ -81,6 +85,138 @@ def test_morpheme_forms(self):
...
+        m = self.tokenizer_obj.tokenize('特Ａ東京')[0]
+        nf = m.normalized_form_morpheme()
+
+        self.assertEqual(m.surface(), '特Ａ')
+        self.assertEqual(m.normalized_form(), '特A')
+        self.assertEqual(nf.word_id(), m.word_id())
+        self.assertEqual(nf.raw_surface(), m.raw_surface())
+        self.assertEqual(nf.surface(), m.surface())
+
+    def test_form_morpheme_with_field_subset(self):
```
hayashi-mas-wap: When a InfoSubset is given, fields not specified in it should not be used (behavior is undefined).

For the normalized/dictionary form morpheme, it is enough if it have a fields specified in the subset (+ headword).

python/src/morpheme.rs
```diff
@@ -248,18 +253,25 @@ impl<'py> Deref for MorphemeRef<'py> {
 pub struct PyMorpheme {
```
hayashi-mas-wap, line 342: I prefer to make this an enum that can be a `Morpheme` or `SingleMorpheme`.
It will be nice to use a `SingleMorpheme` for normalized/dictionary form morpheme rather than a single-element  MorphemeList.

We may also want to make `PyMorphemeList` an enum of a current morpheme list and a list of `SingleMorpheme`.

hayashi-mas-wap (changes requested):

gulldan:
Updated in the latest commits.

- `PyMorpheme` is now backed by either a list-backed `Morpheme` or a real `SingleMorpheme`.
- `PyMorphemeListWrapper` is now backed by either a regular `MorphemeList` or `Vec<SingleMorpheme>`.
- `dictionary_form_morpheme()` / `normalized_form_morpheme()` now return a `SingleMorpheme`-backed Python `Morpheme` for distinct referenced entries, instead of wrapping through a one-entry `MorphemeList`.
- form morphemes now preserve the source `InfoSubset + HEADWORD`.
- standalone split follows Java-style empty / single / multi split behavior, including preserving the parent span for a single replacement split.
- single-backed split `out=` handling now uses the same `try_borrow_mut` error path as list-backed split.

sudachi/src/analysis/mlist.rs
```diff
@@ -218,6 +219,89 @@ impl<D: DictionaryAccess> MorphemeList<D> {
...
+        self.reset_with_word_ids(std::iter::once(word_id), subset)
+    }
+
+    /// Resets this list to dictionary entries for the given word IDs.
+    ///
+    /// This resolves exact dictionary entries instead of looking them up by
+    /// surface. The list surface is the concatenation of dictionary headwords,
+    /// and each entry receives offsets over that standalone surface.
+    #[allow(clippy::result_large_err)]
+    pub fn reset_with_word_ids<I>(&mut self, word_ids: I, subset: InfoSubset) -> SudachiResult<()>
```
hayashi-mas-wap: This method (and `reset_with_word_id`) seems not used now.

sudachi/tests/morpheme.rs
```diff
@@ -32,6 +65,262 @@ fn empty_morpheme_list() {
...
+    assert_eq!(0, splits[0].begin_c());
+    assert_eq!("東京".chars().count(), splits[0].end_c());
+    assert_eq!("東京".chars().count(), splits[1].begin_c());
+    assert_eq!("東京都".chars().count(), splits[1].end_c());
+    assert_eq!("", splits[0].user_data());
+    assert_eq!("", splits[1].user_data());
+}
+
+#[test]
+fn single_morpheme_split_loads_split_subset_on_demand() {
```
hayashi-mas-wap: Please remove this test for now.
As descried in `python/docs/source/topics/subsetting.rst`, the behavior for using a field not included in the passed subset is undefined. We may support an on-demand loading, but don't want to restrict the behavior by the test.

python/tests/test_morpheme.py
```diff
@@ -81,6 +85,205 @@ def test_morpheme_forms(self):
...
+
+            tokenizer = Dictionary(config_path, resource_dir).create()
+            m = tokenizer.tokenize('首都旧')[0]
+            nf = m.normalized_form_morpheme()
+            splits = nf.split(SplitMode.A, add_single=True)
+            out = self.tokenizer_obj.tokenize('東京都', SplitMode.C)[0].split(
+                SplitMode.A)
+            out_result = nf.split(SplitMode.A, out=out, add_single=True)
+            surface_tokenizer = Dictionary(config_path, resource_dir).create(
+                fields={'surface'})
```
hayashi-mas-wap: This `fields` should have `normalized_form` and `split_a` as we access them later.

python/tests/test_morpheme.py
```diff
@@ -81,6 +85,205 @@ def test_morpheme_forms(self):
...
+        nf = m.normalized_form_morpheme()
+
+        self.assertEqual(m.surface(), '特Ａ')
+        self.assertEqual(m.normalized_form(), '特A')
+        self.assertEqual(nf.word_id(), m.word_id())
+        self.assertEqual(nf.raw_surface(), m.raw_surface())
+        self.assertEqual(nf.surface(), m.surface())
+
+    def test_form_morpheme_with_surface_subset(self):
+        tokenizer = self.dict_.create(fields={'surface'})
```
hayashi-mas-wap: This fields should have `dictionary_form` and `normalized_form` as we accesse them later.

python/tests/test_morpheme.py
```diff
@@ -81,6 +85,205 @@ def test_morpheme_forms(self):
...
+        df = m.dictionary_form_morpheme()
+        nf = m.normalized_form_morpheme()
+
+        self.assertEqual(df.surface(), '行く')
+        self.assertEqual(nf.surface(), '行く')
+        self.assertEqual(df.raw_surface(), '行く')
+        self.assertEqual(nf.raw_surface(), '行く')
+
+    def test_form_morpheme_with_reading_subset(self):
+        tokenizer = self.dict_.create(fields={'surface', 'reading_form'})
```
hayashi-mas-wap: This `fields` should have `dictionary_form` and `normalized_form` as we access them later.

python/tests/test_morpheme.py
```diff
@@ -81,6 +85,205 @@ def test_morpheme_forms(self):
...
+        df = m.dictionary_form_morpheme()
+        nf = m.normalized_form_morpheme()
+
+        self.assertEqual(df.surface(), '行く')
+        self.assertEqual(nf.surface(), '行く')
+        self.assertEqual(df.reading_form(), 'イク')
+        self.assertEqual(nf.reading_form(), 'イク')
+
+    def test_single_backed_form_morpheme_uses_projection(self):
+        tokenizer = self.dict_.create(projection='reading')
```
hayashi-mas-wap: An arg `fields` with `dictionary_form` should be set as we access them later.

python/tests/test_morpheme.py
```diff
@@ -81,6 +85,205 @@ def test_morpheme_forms(self):
...
+            config.pop('userDict', None)
+            config.pop('path', None)
+            with open(config_path, 'w', encoding='utf-8') as out:
+                json.dump(config, out, ensure_ascii=False)
+
+            tokenizer = Dictionary(config_path, resource_dir).create()
+            m = tokenizer.tokenize('首都旧')[0]
+            nf = m.normalized_form_morpheme()
+            splits = nf.split(SplitMode.A, add_single=True)
+            out = self.tokenizer_obj.tokenize('東京都', SplitMode.C)[0].split(
```
hayashi-mas-wap: the contents of `out` seems not used. I guess `out = self.tokenizer_obj.tokenize("")` is enough here.

hayashi-mas-wap (changes requested):

gulldan:
agreed, done

hayashi-mas-wap (approved):
LGTM, thank you very much!

PR #336, opened 2026-05-15
Add dictionary entry iteration and lookup_all_entries

### LINKED ISSUE #320: iterate over entries in the dictionary
relates to 
https://github.com/WorksApplications/Sudachi/issues/244
https://github.com/WorksApplications/Sudachi/issues/253

Add method to iterate all (visible / include hidden) entries.
We may need to modify `morpheme` class.

gulldan:
## Summary

Adds dictionary-entry iteration and full-entry lookup APIs for Rust and Python, following Java `develop-v0.8` behavior for `Dictionary.entries()` and `lookupAllEntries()`.

- `entries()` iterates public lexicon entries as standalone morphemes, including entries referred from other entries such as split/constituent units even when they are not indexed for normal lookup.
- Internal entries automatically generated for literal normalized forms are not exposed.
- `lookup_all_entries()` normalizes the query with dictionary input-text plugins, scans public lexicon entries, and returns entries whose normalized surface matches.
- Indexed `lookup()` also normalizes the query before trie lookup and now materializes results over the normalized input text, matching Java behavior for cases such as `IgnoreYomigana`.

## Review notes

- WordInfo entry scanning is now fallible, and dictionary load validates entry boundaries. Malformed or truncated entry blocks no longer silently truncate `entries()` / `lookup_all_entries()`.
- Entry-scan helpers stay crate-internal; the public API surface is the dictionary methods and the Python binding methods.
- The iteration order is intentionally not part of the public contract.

## Validation

- `cargo fmt --all --check`
- `cargo test -p sudachi dic::word_info::infos --offline`
- `cargo test -p sudachi --test dictionary --offline`
- `cargo test -p sudachipy --lib --offline`
- `python setup.py build_rust --inplace && python -m unittest tests.test_dictionary`
- `git diff --check`

## Benchmark note

Checked on SudachiDict full 20260428, built as a V1 dictionary from raw CSV. Java comparison used `WorksApplications/Sudachi` `develop-v0.8` at `9a2f4c3`. Queries: `東京都`, `京都`, `京都（キョウト）`, `外国人参政権`, `存在しない語`. Rust and Python were measured with release builds; RSS was measured with `/usr/bin/time -l`.

All engines matched for `entries()`, indexed `lookup()`, and `lookup_all_entries()` digests. `entries()` count was `2,883,177` with checksum `sum=0x79401354812f1721`, `xor=0xc66f60b830cc6537`.

| Engine | `entries()` | indexed `lookup` | `lookup_all_entries` | peak RSS |
|---|---:|---:|---:|---:|
| Rust | 1.533 s | 1.061 us/call | 1.304 s/call | 172 MB |
| Java | 1.456 s | 7.475 us/call | 0.471 s/call | 5012 MB |
| Python | 27.644 s | 1.248 us/call | 1.273 s/call | 191 MB |

`lookup_all_entries("京都")` and `lookup_all_entries("京都（キョウト）")` both returned 2 entries, while indexed `lookup(...)` returned 1 entry for each. The Python `entries()` timing includes Python-level iteration and method-call overhead over 2.88M morphemes; the core scan path is exercised by Rust and by Python `lookup_all_entries()`.

Closes #320

### DIFF
--- a/sudachi/src/dic/word_info/infos.rs
+++ b/sudachi/src/dic/word_info/infos.rs
@@ -40,32 +40,19 @@ impl<'a> WordInfos<'a> {
     }
 
     pub fn entry_ids_in_order(&self, num_total_entries: u32) -> Option<Vec<EntryId>> {
-        let mut result = Vec::with_capacity(num_total_entries as usize);
-        let mut offset = Self::ENTRIES_INITIAL_OFFSET;
-
-        while result.len() < num_total_entries as usize {
-            if offset % Self::WORD_INFO_OFFSET_ALIGNMENT != 0 {
-                return None;
-            }
-
-            let entry_id = EntryId::new((offset >> Self::WORD_ID_ALIGNMENT_BITS) as u32);
-            result.push(entry_id);
-
-            let size = self.entry_size_at(offset)?;
-            offset = offset.checked_add(size)?;
-        }
-
-        Some(result)
+        self.entry_ids(num_total_entries)
+            .collect::<SudachiResult<Vec<_>>>()
+            .ok()
     }
 
-    pub fn entry_ids(&self, num_total_entries: u32) -> WordInfoEntryIdIter<'_, '_> {
+    pub(crate) fn entry_ids(&self, num_total_entries: u32) -> WordInfoEntryIdIter<'_, '_> {
         WordInfoEntryIdIter {
             infos: self,
             cursor: Self::entry_id_cursor(num_total_entries),
         }
     }
 
-    pub fn entry_id_cursor(num_total_entries: u32) -> WordInfoEntryIdCursor {
+    pub(crate) fn entry_id_cursor(num_total_entries: u32) -> WordInfoEntryIdCursor {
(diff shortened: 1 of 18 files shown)


sudachi/src/dic/word_info/infos.rs
```diff
@@ -109,6 +152,45 @@ impl<'a> WordInfos<'a> {
     }
 }
 
+fn invalid_entry_block(offset: usize, reason: &str) -> SudachiError {
```
hayashi-mas-wap: Let's make a new error for this.
maybe `word_info/infos.rs::WordInfoError`, with `EntryIdNotAligned(offset)`, `FailedToLoadEntrySize(offset)` and `EntryIdTooLarge(offset)`.

hayashi-mas-wap (approved):
LGTM!
I want to request to define new error enum.

sudachi/src/dic/word_info/infos.rs
gulldan: done

PR #342, opened 2026-05-23
Add Python user_data access

### LINKED ISSUE #338: Add user data access for python
user_data field of a morpheme is not accessible with python API

gulldan:
## Summary
- Expose `Morpheme.user_data()` in the Python bindings for both analyzed morphemes and standalone dictionary-entry morphemes.
- Add `user_data` to Python field subsetting, type stubs, and subsetting docs, including an example for partial loading.
- Also fix stale subsetting docs that referred to `splits_a` / `splits_b`; accepted field names are `split_a` / `split_b`.
- Cover user-data access from tokenization, `Dictionary.lookup()`, `Dictionary.entries()`, `lookup_all_entries()`, and dictionaries built from user CSVs.

## Root cause
The v0.7 Rust core already stores and resolves `user_data`, but the Python `Morpheme` API and Python field-subset parser did not expose or load that field.

## Validation
- `cargo fmt --all`
- `git diff --check`
- `cargo test -p sudachi user_data`
- `cargo test -p sudachipy`
- `(cd python && python -m pip install -e .)`
- `python -m unittest python.tests.test_build python.tests.test_morpheme python.tests.test_dictionary python.tests.test_tokenizer`
- `python -m unittest python.tests.test_dictionary python.tests.test_morpheme`
- `cargo test --workspace`
- `python -m unittest discover python/tests`

Closes #338

### DIFF
--- a/python/tests/test_morpheme.py
+++ b/python/tests/test_morpheme.py
@@ -329,10 +329,6 @@ class TestTokenizer(unittest.TestCase):
         m = tokenizer.tokenize('すだち')[0]
         self.assertEqual(m.user_data(), '徳島県産')
 
-        tokenizer = self.dict_.create(fields=set())
-        m = tokenizer.tokenize('すだち')[0]
-        self.assertEqual(m.user_data(), '')
-
     def test_normalize_half_full(self):
         m = self.tokenizer_obj.tokenize('特Ａ東京')
         self.assertEqual(len(m), 2)
--- a/python/tests/test_morpheme.py
+++ b/python/tests/test_morpheme.py
@@ -318,6 +318,21 @@ class TestTokenizer(unittest.TestCase):
         m = self.tokenizer_obj.tokenize('東京府')[0]
         self.assertEqual(m.synonym_group_ids(), [1, 3])
 
+    def test_morpheme_user_data(self):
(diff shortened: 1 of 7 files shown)


python/tests/test_morpheme.py
```diff
@@ -318,6 +318,21 @@ def test_morpheme_synonym_group_ids(self):
...
+        self.assertEqual(m.user_data(), '徳島県産')
+
+        m = self.tokenizer_obj.tokenize('京都')[0]
+        self.assertEqual(m.user_data(), '')
+
+        tokenizer = self.dict_.create(fields={'user_data'})
+        m = tokenizer.tokenize('すだち')[0]
+        self.assertEqual(m.user_data(), '徳島県産')
+
+        tokenizer = self.dict_.create(fields=set())
```
hayashi-mas-wap: This part should be omitted, as we don't define the behavior for accessing non-specified fields.

hayashi-mas-wap (changes requested):

python/tests/test_morpheme.py
gulldan: done

hayashi-mas-wap (approved):

Issue #117, opened 2021-10-20
Explore prefetch-compatible trie layout


eiennohito:
When building a dictionary, we need to build Trie index.

Probably, we shouldn't move away from [darts-clone](https://github.com/s-yata/darts-clone)-compatible structure, but there can be variations on the construction. It would be interesting to explore whether it is possible to make prefetch-compatible trie that could load the trie entries in L1 cache for the next symbol speculatively, even with not 100% accuracy (e.g. getting hits for high-frequency entries).

Rust Implementations to Note: https://github.com/takuyaa/yada (it seems to follow darts-clone internal structure with original darts construction procedure)

hayashi-mas-wap:
I add some context based on my understanding.
- The construction of a double array binary representation has some degree of freedom:
    - how to order the nodes of TRIE
    - how to search offset for each node's edge array
- With a statistical analysis of the large corpus, we may able to utilize this to make the binary prefetch-friendly
    - e.g. layout nodes closely if they are frequently accessed consecutively

eiennohito:
IMO layout should not be aim to put frequently used close together (which is probably impossible), but similarly (=having mostly same relative offsets, e.g. subarrays for each utf-8 japanese triplet should have more or less similar offsets relative to each other, modulo their values). 

Because natural languages follow Zipf law, laying out even few entries this way can get measurable effect (I think)

PR #347, opened 2026-05-29
Issue #117: runtime prefetch-friendly trie lookup (+~9%, format-preserving)

gulldan:
## What

A **runtime** answer to #117: a software-pipelined, prefetch-assisted common-prefix search. It is **on by default**, produces **byte-for-byte identical** output, and is **+8.5–9.5% faster end-to-end** on real SudachiDict small/core/full — with **no change to the serialized dictionary format** and **zero build-time cost**.

## How

Sudachi runs one independent trie walk per character boundary, and each walk's next node address is known one step ahead. So `K = 4` walks are kept in flight and each lane's next node is `prefetch`ed while the others run (group prefetching / AMAC). New `Trie::common_prefix_batch` + `Lexicon`/`LexiconSet::lookup_batch`, wired into `build_lattice` while preserving the reachability gate, OOV handling, the `can_bow` filter, and lookup order. A shared `step_once` makes the scalar and pipelined paths identical by construction. Portable hint: `_mm_prefetch` (x86) / `prfm pldl1keep` (aarch64) / no-op elsewhere — stable Rust only.

## Measured (Apple M4 Max, median of 31 runs, CV < 2%)

| tier | scalar | pipelined | speedup |
|---|---:|---:|---:|
| small | 182.8 ms | 168.3 ms | **+8.6%** |
| core | 189.6 ms | 174.8 ms | **+8.5%** |
| full | 197.5 ms | 180.5 ms | **+9.5%** |

Parity: unit tests across lane counts {1..16} + a differential tokenization test (modes A/B/C) + the full existing suite, all green on the pipelined path.

## Also in this PR

- `docs/performance-investigation.md` — a measured study including **negative results**: prefetching the connection matrix *hurts* (it is volume-bound, not miss-bound), and an alternative trie backend (Crawdad/Daachorse) is only ~+3–5% end-to-end for a dictionary-format break.
- Batch parallelism scales **~8.9× on 16 cores** (the dictionary is `Send + Sync`, shared via `Arc`); see `examples/tokenize_parallel_bench`.
- The earlier **build-time** node relayout (`--trie-layout cache-aware`) is kept strictly **opt-in** (~14× build time for ≤1.10× lookup — not default-worthy).
- The cross-method comparison crates (daachorse/crawdad/fst/rsmarisa) are gated behind `--features matcher-comparison`, so they stay out of the default build/CI graph.

Benchmarks reproduce with `examples/tokenize_pipeline_bench` (single-thread A/B, reports median/CV) and `examples/tokenize_parallel_bench` (scaling) — no extra dependencies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,6 +207,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 
 [[package]]
+name = "crawdad"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9742b11612a5e342c05ae62386a7a486c8fcf60bf835914a469d44c545c3b7fe"
+
+[[package]]
 name = "criterion"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -295,10 +301,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "daachorse"
+version = "3.0.0"
+source = "git+https://github.com/daac-tools/daachorse?tag=v3.0.0#41755e780a41b49f041e76f3d1e81bd0129188c7"
+
+[[package]]
 name = "default_input_text"
 version = "0.6.11-a1"
 dependencies = [
  "sudachi",
+]
+
+[[package]]
+name = "deranged"
+version = "0.5.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7cd812cc2bc1d69d4764bd80df88b4317eaef9e773c75226407d9bc0876b211c"
+dependencies = [
+ "powerfmt",
 ]
 
 [[package]]
@@ -345,6 +365,12 @@ name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
+
+[[package]]
+name = "fst"
+version = "0.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7ab85b9b05e3978cc9a9cf8fea7f01b494e1a09ed3037e16ba39edc7a29eb61a"
 
 [[package]]
 name = "getrandom"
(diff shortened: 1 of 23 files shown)


gulldan:
Both the construction-side idea described here and a runtime variant were
  implemented and benchmarked in #347, on real SudachiDict small/core/full + the
  full `matrix.def`, 16k Kyoto-lead sentences (Apple M4 Max, `--release`).

  ### 1. The build-time idea from this issue — tested

  `--trie-layout cache-aware`: a darts-clone/Yada-compatible builder with
  hot-first recursion and cache-line-aware offset placement, biased toward
  high-frequency entries via three weightings — uniform, dictionary prefix-count,
  and an external frequency profile (`surface<TAB>count`). It emits the exact same
  `u32` double-array format and is readable by the current runtime.

  Lookup speedup vs the classic builder (common-prefix at every char start):

  | tier | lookup speedup | build time |
  |---|---|---|
  | small | 1.00–1.05× | ~6 s → ~60 s |
  | core | 1.01–1.05× | ~14 s → ~190 s |
  | full | 1.07–1.10× | ~41 s → ~580 s |

  So the frequency-biased layout does give a small win, but it only reaches the
  5–10% range on `full`, costs ~14× build time, and grows the trie ~1–2%. Kept
  **opt-in**, not the default.

  ### 2. Runtime software prefetch — what actually pays off

  Same goal (speculatively pull the next symbol's node into L1, imperfect accuracy
  is fine), but at lookup time instead of via layout: the lattice runs one
  independent common-prefix walk per character boundary, so K=4 walks are kept in
  flight and each lane's next node is `prefetch`ed while the others run
  (group-prefetching / AMAC). Format and build are unchanged.

  **+8.5–9.5% end-to-end tokenization** (median of 31 runs, CV < 2%), byte-for-byte
  identical output, zero build/size/memory cost.

  @eiennohito Could you clarify the target you had in mind?

  - Is a runtime-prefetch approach acceptable, or did you specifically want the
    speedup baked into the serialized layout / construction?
  - What lookup (or end-to-end) speedup would justify changing the default?
  - Must output stay exactly identical, or is an approximate path on the table?
    Profiling shows the Viterbi connection-matrix step (~23%), not the trie (~17%),
    is the dominant cost; a larger win there needs beam-style pruning, which is
    approximate.

  Full write-up incl. negative results: `docs/performance-investigation.md` and
  `docs/trie-prefetch.md` in #347.

gulldan:
Superseded by #348, a focused PR with only the runtime prefetch + varint decoder (no build-time cache-aware experiment, no cross-method comparison deps). The broader exploration (build-time relayout, multithreading scaling, connection-ID remap study) stays on the `feat/dict-layout-117` branch and can become a separate follow-up if useful.

PR #348, opened 2026-05-29
Issue #117: runtime prefetch-friendly trie lookup

gulldan:
## What

A **runtime** answer to #117: a software-pipelined, prefetch-assisted
common-prefix search, plus a faster word-id varint decoder. On by default,
**byte-for-byte identical output**, **~+12% end-to-end** tokenization on real
SudachiDict tiers, with **no change to the serialized dictionary format** and
**zero build-time cost**.

## How

- **Prefetch pipeline.** Sudachi runs one independent trie walk per character
  boundary, and each walk's next node address is known one step ahead. So `K = 4`
  walks are kept in flight and each lane's next node is `prefetch`ed while the
  others run (group prefetching / AMAC). New `Trie::common_prefix_batch` +
  `Lexicon`/`LexiconSet::lookup_batch`, wired into `build_lattice` while
  preserving the reachability gate, OOV handling, the `can_bow` filter, and
  lookup order. A shared `step_once` makes the scalar and pipelined paths
  identical by construction. Portable hint: `_mm_prefetch` (x86) /
  `prfm pldl1keep` (aarch64) / no-op elsewhere — stable Rust only.
- **Faster word-id decode.** The word-id table iterator now uses a direct
  `decode_varint32` with a single-byte fast path instead of routing each delta
  through the nom `varint32`/`varint64` (per-byte `Result`, `u64` widening).

## Measured (Apple M4 Max, `--release`, median of 31 runs, CV < 2%)

Prefetch pipeline, end-to-end tokenization of 16k Kyoto-lead sentences:

| tier | scalar | pipelined | speedup |
|---|---:|---:|---:|
| small | 182.8 ms | 168.3 ms | **+8.6%** |
| core | 189.6 ms | 174.8 ms | **+8.5%** |
| full | 197.5 ms | 180.5 ms | **+9.5%** |

The direct varint decoder adds a further **+2.7%** (interleaved A/B on the full
dict) on top, for **~+12%** combined.

## Correctness

- Unit tests: `common_prefix_batch` matches `common_prefix_iterator` across lane
  counts {1..16}, prefetch on/off, and edge cases; `decode_varint32` matches the
  nom decoder including `u32::MAX`.
- `tests/pipelined_lookup_parity.rs`: full tokenization (modes A/B/C) is
  identical between the scalar and pipelined lattice builders.
- The whole existing test suite passes on the pipelined path (the default).

Reproduce the numbers with `examples/tokenize_pipeline_bench` (interleaves the
scalar and pipelined paths, reports median/CV).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gulldan:
I want to make sure I fully understand what's needed to resolve this before moving forward. Could you clarify the expected outcome, or point me to any relevant context? Once I know which direction to take, I'll get it sorted and close out the issue.

hayashi-mas-wap:
@eiennohito Do you have comment about this?
I think this PR is matching to the idea you wrote in #117.

eiennohito:
@hayashi-mas-wap 
I think the idea is interesting, but is a halfway there. To exploit this idea in full, lattice boundary node creation should be done after the trie walk, not interleaved with it (as those compete for L1 cache).

My original idea was to take a largish corpus and try to layout the trie itself so it will be more prefetchable (few steps at a time). Together with this approach there can be even bigger winnings and if the patterns will be more exploitable by hw prefetches, even Java version should have free wins.

But free performance gains are always welcome.

As a rant from me this looks a bit too much Claude Code output (nothing against it) without any human filtering/cleanup afterwards (super against this).

gulldan:
@eiennohito thank you for taking a look and for clarifying the original intent.

I agree that this PR is not the full answer to #117 yet. I treated it as an R&D / PoC step first, because I wanted to understand whether a runtime-only approach can provide any useful format-preserving gain before moving into a deeper trie-layout experiment.

I also did some additional research: I checked related approaches, alternative trie implementations, similar libraries, and relevant literature, then summarized the findings in a separate note. The goal was to make the trade-offs and negative results explicit, not just push code.

The current implementation is still draft-quality. It was written to validate the idea and collect measurements, so I fully expect it to need cleanup and further review. If this direction is acceptable, I will revise the code and documentation to make it much cleaner and more aligned with the project.

From what I understand now, this PR can be considered a small runtime optimization, while the original corpus/layout-based trie idea may need a separate follow-up investigation.
 
wall

eiennohito:
Please do not post Claude Code walls of text if you do not understand what they say. MOST of your analysis is incorrect.

I think that Claude Code misattributed the DRAM access latency (which trashes all the cache) during the analysis.
Connection matrix access happens concurrently with WordInfo parsing from the binary dictionary (wrt in-CPU instruction reordering and OoO execution) and DRAM backend stalls can be attributed to incorrect places. Claude Code by default misattributes results of `perf annotate`. You need to guide it to have correct conclusions.

The correct fix for connection matrix is to relearn it from scratch (the best) or simplify the current one. If you do not know, the connection matrix in MeCab style analyzers are CRF feature weights, summed up per binary feature templates and quantized to i16 values. It is possible to modify CRF to make the feature learning more smart (e.g. L1-igh regularization), so the noise from weird features can be skipped.

hayashi-mas-wap:
@eiennohito 
Thank you for taking the time on this.
I've added a supplementary comment on #117.

@gulldan 
The scope of the comment above is too large for a single PR.
While performance improvement is of course welcome, please create issues for each implementation ideas (or an issue to discuss) rather than putting them here.

As the idea of your commits so far are independent from #117, let's focus on polishing the current implementation in this PR.

gulldan:
@eiennohito
thank you for detailed answer.
you are absolutely right

@hayashi-mas-wap 
thank you ill do it on weekends

gulldan:
Cleaned this up. The PR is now only the prefetch lookup: the pipelined trie search plus its lattice wiring, nothing else. I took out the varint decoder, the WordInfo cache and the notes/benchmark, they didn't belong here.

Output is identical to develop-v0.7 (diffed the full kyoto-leads corpus, 16k sentences) and the tests pass. About 10% faster on do_tokenize on the full dict (median 1.10x over 31 interleaved scalar/pipelined trials).

hayashi-mas-wap (approved):

PR #349, opened 2026-06-05
Add Python TextNormalizer API

### LINKED ISSUE #339: Add TextNormalizer for python
Add [TextNormalizer](https://github.com/WorksApplications/Sudachi/blob/develop-v0.8/src/main/java/com/worksap/nlp/sudachi/TextNormalizer.java) class for python.

We may also want to add method to create text normalizer from dictionary (which use same `InputTextPlugin` set).

gulldan:
## Summary
- Add `sudachipy.TextNormalizer` with `normalize(text)`.
- Add `Dictionary.text_normalizer()` so normalization uses the same dictionary input-text plugin set as tokenization.
- Export the API in Python, update type stubs, docs, changelog, and tests.

## Maintainer note
This follows the Java `TextNormalizer` shape from `develop-v0.8`: the dictionary-backed constructor path reuses the dictionary grammar and configured input-text plugins, so custom plugins such as `IgnoreYomiganaPlugin` are reflected in `normalize()`.

`TextNormalizer` normalizes raw input text through input-text plugins. It does not perform morphological analysis or return the dictionary-normalized forms exposed by `Morpheme.normalized_form()`.

This PR targets `develop-v0.7`, so I am using a reference instead of a closing keyword for the issue link.

Refs #339.

## Validation
- `cargo fmt --all --check`
- `cargo test --workspace`
- `.env-py3/bin/python -m pip install -e '.[tests]'`
- `.env-py3/bin/python -m unittest`
- `PATH="$PWD/.env-py3/bin:$PATH" bash build_and_test.sh`
- `.env-py3/bin/python -m py_compile py_src/sudachipy/__init__.py py_src/sudachipy/sudachipy.pyi tests/test_import.py tests/test_text_normalizer.py`
- `cargo test -p sudachipy`
- `git diff --check`

### DIFF
--- a/python/src/text_normalizer.rs
+++ b/python/src/text_normalizer.rs
@@ -21,42 +21,65 @@ use pyo3::prelude::*;
 use sudachi::dic::DictionaryAccess;
 use sudachi::error::SudachiResult;
 use sudachi::input_text::InputBuffer;
+use sudachi::text_normalizer::TextNormalizer;
 
 use crate::dictionary::{PyDicData, PyDictionary};
 use crate::errors;
 
 /// A text normalizer.
 ///
-/// This applies dictionary input-text plugins to raw input text. It does not
-/// perform morphological analysis or return morpheme normalized forms.
+/// This applies input-text plugins to raw input text. It does not perform
+/// morphological analysis or return morpheme normalized forms.
 ///
 /// Create using ``Dictionary.text_normalizer()`` or by passing a
-/// ``Dictionary`` to this class.
+/// ``Dictionary`` to this class. Without a dictionary, this uses the default
+/// input-text normalization.
 #[pyclass(module = "sudachipy", name = "TextNormalizer")]
 pub struct PyTextNormalizer {
-    dict: Arc<PyDicData>,
+    inner: PyTextNormalizerInner,
+}
+
+enum PyTextNormalizerInner {
+    Default(TextNormalizer<'static>),
+    Dictionary {
+        dict: Arc<PyDicData>,
+        buffer: InputBuffer,
+    },
 }
 
 impl PyTextNormalizer {
     pub(crate) fn from_dictionary(dict: Arc<PyDicData>) -> Self {
-        Self { dict }
+        Self {
+            inner: PyTextNormalizerInner::Dictionary {
+                dict,
+                buffer: InputBuffer::new(),
+            },
+        }
     }
 }
 
 #[pymethods]
 impl PyTextNormalizer {
-    /// Creates a text normalizer from a dictionary.
+    /// Creates a text normalizer.
     ///
-    /// The normalizer applies the same input-text plugins that the dictionary
-    /// uses before tokenization. The normalizer can keep normalizing text after
-    /// the source dictionary is closed.
+    /// When dictionary is provided, this applies the same input-text plugins
+    /// used by that dictionary before tokenization. Without a dictionary, this
+    /// applies the default input-text normalization.
     #[new]
-    #[pyo3(text_signature = "(dictionary) -> TextNormalizer")]
-    fn new(dictionary: PyRef<'_, PyDictionary>) -> PyTextNormalizer {
-        PyTextNormalizer::from_dictionary(dictionary.data())
+    #[pyo3(
+        signature = (dictionary=None),
+        text_signature = "(dictionary=None) -> TextNormalizer"
+    )]
+    fn new(dictionary: Option<PyRef<'_, PyDictionary>>) -> PyResult<PyTextNormalizer> {
+        match dictionary {
+            Some(dictionary) => Ok(PyTextNormalizer::from_dictionary(dictionary.data())),
+            None => Ok(Self {
+                inner: PyTextNormalizerInner::Default(errors::wrap(TextNormalizer::default())?),
+            }),
+        }
     }
 
-    /// Normalize text using dictionary input-text plugins.
+    /// Normalize text using input-text plugins.
     ///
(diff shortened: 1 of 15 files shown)


python/src/text_normalizer.rs
```diff
@@ -0,0 +1,83 @@
...
+/// A text normalizer.
+///
+/// This applies dictionary input-text plugins to raw input text. It does not
+/// perform morphological analysis or return morpheme normalized forms.
+///
+/// Create using ``Dictionary.text_normalizer()`` or by passing a
+/// ``Dictionary`` to this class.
+#[pyclass(module = "sudachipy", name = "TextNormalizer")]
+pub struct PyTextNormalizer {
+    dict: Arc<PyDicData>,
```
hayashi-mas-wap: I also want a "default" text normalizer, which corresponds to the current rust side text normalizer.
It only needs `DefaultInputPlugin` and its setup requires config with embedded resource and empty grammar.

we may need to make this enum to achieve that.

python/src/text_normalizer.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn normalize(&self, py: Python<'_>, text: &str) -> PyResult<String> {
+        errors::wrap_ctx(
+            py.detach(|| normalize_with_dictionary(&self.dict, text)),
+            "Error during text normalization",
+        )
+    }
+}
+
+fn normalize_with_dictionary(dict: &PyDicData, text: &str) -> SudachiResult<String> {
+    let mut buffer = InputBuffer::new();
```
hayashi-mas-wap: Although current rust text-normalizer also instantiates `InputBuffer` for each call of `normalize`, it would be better to keep and reuse it like `StatefulTokenizer` does.

python/tests/test_text_normalizer.py
```diff
@@ -0,0 +1,94 @@
...
+        normalizer = TextNormalizer(self.dict_)
+        self.assertEqual("abc", normalizer.normalize("ＡＢＣ"))
+
+    def test_repeated_calls_and_empty_text(self):
+        normalizer = self.dict_.text_normalizer()
+        self.assertEqual("", normalizer.normalize(""))
+        self.assertEqual("abc", normalizer.normalize("ＡＢＣ"))
+        self.assertEqual(NORMALIZED_TEXT, normalizer.normalize(ORIGINAL_TEXT))
+
+    def test_normalizer_outlives_dictionary_close(self):
```
hayashi-mas-wap: I don't want to test this behavior here. This depends too much on how we keep the dictionary inside the normalizer (same for the doc string in the sudachipy.pyi etc.).

hayashi-mas-wap (changes requested):
Thank you for the PR, I wrote some comments.
If possible, I would be great if you also update rust side text-normalizer to be built from a dictionary.

gulldan:
done

sudachi/src/text_normalizer.rs
```diff
@@ -14,33 +14,216 @@
...
+/// By default, this uses `DefaultInputTextPlugin`. When built from a dictionary,
+/// it applies that dictionary's configured input-text plugins.
+pub struct TextNormalizer<'a> {
+    source: TextNormalizerSource<'a>,
+    input: InputBuffer,
+}
+
+enum TextNormalizerSource<'a> {
+    Default(DefaultInputTextPlugin),
+    BorrowedDictionary(&'a (dyn DictionaryAccess + Sync)),
```
hayashi-mas-wap: `BorrowedDictionary` and `SharedDictionary` can be merged into single enum variant which has `D: DictionaryAccess` as the `StatefulTokenizer` keeps `dict`.

sudachi/src/text_normalizer.rs
```diff
@@ -14,33 +14,216 @@
...
+    let cfg = ConfigBuilder::empty().push_embedded().build();
+    plugin.set_up(
+        &serde_json::Value::Object(serde_json::Map::default()),
+        &cfg,
+        grammar,
+    )?;
+    Ok(plugin)
+}
+
+fn empty_grammar() -> SudachiResult<Grammar<'static>> {
```
hayashi-mas-wap: Let's move these to `ConnectionMatrix::empty` and `Grammar::empty`.

sudachi/src/text_normalizer.rs
```diff
@@ -14,33 +14,216 @@
...
+            &self.oov_provider_plugins
+        }
+
+        fn path_rewrite_plugins(&self) -> &[Box<dyn PathRewritePlugin + Sync + Send>] {
+            &self.path_rewrite_plugins
+        }
+    }
+
+    #[test]
+    fn default_normalizer_reuses_buffer() {
```
hayashi-mas-wap: This test name looks unnatural for me, maybe just `default_normalizer_works` or something?

hayashi-mas-wap (changes requested):

gulldan:
@hayashi-mas-wap  thank you for review

  1. removed the core `TextNormalizerSource` enum and made `TextNormalizer` generic over its source type: `TextNormalizer<D = DefaultInputTextPlugin>`.

     This keeps the dictionary directly in the normalizer, the same way `StatefulTokenizer` keeps `dict`. As a result, borrowed dictionaries and shared dictionaries do not need separate enum variants anymore: both are just concrete `D:  DictionaryAccess` sources.

2.  added `ConnectionMatrix::empty()` and `Grammar::empty()` and removed the local empty grammar construction from `text_normalizer.rs`
3. test renamed - default_normalizer_works, it more clear

hayashi-mas-wap (approved):

